Issue #1, opened 2023-07-28
Repository template (file layouts, cookiecutter templating engines, distribution options)


agriyakhetarpal:
Starting this as a placeholder issue for tracking down tasks to be completed and those that are complete. I will be dividing these into separate issues and PRs

## Cookiecutters

1. https://github.com/cookieninja-generator/cookieninja
2. https://github.com/cookiecutter/cookiecutter

as suggested by @Saransh-cpp

### Examples
1. https://github.com/osl-incubator/scicookie
2. https://github.com/scientific-python/cookie
4. https://github.com/copier-org/copier
5. https://github.com/cruft/cruft/
6. [Cookiecutter Data Science (drivendata.github.io)](https://drivendata.github.io/cookiecutter-data-science/)

Suggestion from @brosaplanella: "A bit different, but Julia has [DrWatson.jl](https://github.com/JuliaDynamics/DrWatson.jl) which has many cool features, maybe we can get some ideas".

## Possible layout

The folder structure can look like this

```python
├── .github/workflows/
├── src
└── docs
      ├── examples/  # notebooks that can be rendered with nbsphinx
      ├── _static/
      ├── sphinxext/
      ├── source/
      └── conf.py
├── tests/  # (optional)
├── examples/  # alternatively, example scripts or notebooks that are not to be rendered with the Sphinx builder
├── pyproject.toml
├── README.md
├── .pre-commit-config.yaml
├── .readthedocs.yaml
├── .env # contains DATA_PATH
├── noxfile.py  # (or tox.ini, if users want to use tox)
```

## The required documentation should

* Explain how to clone this template to start a new project
* Explain how to rename the project in `pyproject.toml` and `docs/conf.py`
* Explain how to structure a typical project with source files and utility classes and methods (in `src/`), unit tests with
* Parameter entry points in `pyproject.toml`, with
```
[project.entry-points.pybamm_parameter_sets]
MyParameters = "package_name.parameters.my_parameters:get_parameter_values"
```
* point to the PyBaMM documentation, wherein the developments and advancements in this repository shall be reflected on a separate page or section in the user guide

which can then be accessed as `pybamm.ParameterValues(“MyParameters”)` in the source code

### Build-backends

1. `hatch`
2. `flit`
3. `setuptools` (later)

### Documentation

1. Theme
2. Sphinx extensions

### Project structure

1. Project metadata in `pyproject.toml`
2. and so on

agriyakhetarpal:
[...]

PR #2, opened 2023-08-01
Initial draft for a `cookiecutter` template (licenses and folder structure)

agriyakhetarpal:
In specific, the MIT, Apache, and BSD-3-Clause licenses. I am adding them in a PR for visibility and for keeping track of changes more effectively; should we add templates for more licenses?

Edit: this PR also adds an initial structure for the template files to build upon further. The three licenses are adequate since we want to encourage the use of permissive licensing


Saransh-cpp (commented):
I think it would also be nice to add cookiecutter.json and cookiecutter support through this PR, so that we can test this locally? (This looks like that it will work, but would be nice to have `cookiecutter.json` and `pyproject.toml` for the base repository, not the template.)

agriyakhetarpal:
I did have an outline for a template ready to push. The template works and is a more bare-bones rendition of the `scientific-python` one (only `hatch` is supported at the moment).

I know this is now a larger PR for reviewing now (😅); I am happy to write separate, shorter PRs for further tasks such as adding utility functions in https://github.com/pybamm-team/pybamm-cookiecutter/issues/1#issuecomment-1660389638, adding entry points for parameter sets, and more

cookiecutter.json
```diff
@@ -0,0 +1,43 @@
...
+        "full_name": "Your name",
+        "email": "Your email",
+        "project_short_description": "A short description of your project",
+        "license": "Select a license",
+        "backend": {
+            "__prompt__": "Choose a build backend",
+            "hatch": "Hatchling (recommended for pure Python projects)"
+        }
+    }
+}
```
Saransh-cpp, line 43:
```suggestion
}

```

docs/conf.py
```diff
@@ -0,0 +1,45 @@
+import importlib.metadata
+
+project = "pybamm-cookiecutter"
+copyright = "2023, Agriya Khetarpal"
+author = "Agriya Khetarpal"
```
Saransh-cpp, line 5: How about -

```py
project = "PyBaMM"
copyright = "2018-2023, The PyBaMM Team"
author = "The PyBaMM Team"
```

src/pybamm_cookiecutter/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""
+Copyright (c) 2023 Agriya Khetarpal. All rights reserved.
```
Saransh-cpp, line 2: Maybe we can attribute pybamm team here? Includes everyone!

src/pybamm_cookiecutter/__init__.py
```diff
@@ -0,0 +1,10 @@
...
+
+pybamm-cookiecutter: A template for creating battery modeling projects based on PyBaMM
+"""
+
+
+__version__ = "0.1.0"
```
Saransh-cpp, line 8: Let's start with this

```suggestion
__version__ = "0.0.1"
```

{{cookiecutter.project_name}}/.pre-commit-config.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+ci:
+  autoupdate_commit_msg: "chore: update pre-commit hooks"
+  autofix_commit_msg: "style: pre-commit fixes"
+
+repos:
+  - repo: https://github.com/psf/black
+    rev: "23.7.0"
+    hooks:
+      - id: black-jupyter
```
Saransh-cpp, line 13: We already have `ruff`!

```suggestion
```

{{cookiecutter.project_name}}/pyproject.toml
```diff
@@ -0,0 +1,116 @@
...
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",  # PyBaMM supports >=3.8,<3.12
+  "Topic :: Scientific/Engineering",
+]
+{%- if cookiecutter.backend == "hatch" %}
+dynamic = ["version"]
+dependencies = ["pybamm"]
+{%- endif %}
```
Saransh-cpp, line 44:
```suggestion
{%- endif %}
dependencies = ["pybamm"]
```

{{cookiecutter.project_name}}/pyproject.toml
```diff
@@ -0,0 +1,116 @@
...
+Homepage = "{{ cookiecutter.url }}"
+"Bug Tracker" = "{{ cookiecutter.url }}/issues"
+Discussions = "{{ cookiecutter.url }}/discussions"
+Changelog = "{{ cookiecutter.url }}/releases"
+
+[tool.hatch]
+version.path = "src/{{ cookiecutter.__project_slug }}/__init__.py"
+envs.default.dependencies = [
+  "pybamm",
+]
```
Saransh-cpp, line 68:
```suggestion
{%- if cookiecutter.backend == "hatch" %}
[tool.hatch]
version.path = "src/{{ cookiecutter.__project_slug }}/__init__.py"
envs.default.dependencies = [
  "pybamm",
]
{%- endif %}
```

{{cookiecutter.project_name}}/src/{{cookiecutter.__project_slug}}/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""
+Copyright (c) {{ cookiecutter.__year }} {{ cookiecutter.full_name }}. All rights reserved.
+
+{{ cookiecutter.project_name }}: {{ cookiecutter.project_short_description }}
+"""
+
+
+__version__ = "0.1.0"
```
Saransh-cpp, line 8:
```suggestion
__version__ = "0.0.1"
```

src/pybamm_cookiecutter/__init__.py
Saransh-cpp: till the time we switch to VCS versioning ;)

Saransh-cpp (changes requested):
Awesome, thanks @agriyakhetarpal! This is the perfect sized PR, very minimalistic, but something that can be tested.

src/pybamm_cookiecutter/__init__.py
agriyakhetarpal: Fair enough!

agriyakhetarpal:
How should we test whether the cookiecutter works as intended? We could use a context manager along with [tempfile](https://docs.python.org/3/library/tempfile.html) to create cookiecutters with a specification and test whether `pyproject.toml` contains specific fields with `tomllib`, and the version of the created package could be tested upon importing it with `importlib.metadata`. Not sure how other cookiecutters do it

Saransh-cpp (approved):
Thanks @agriyakhetarpal@ Looks good! For the tests, we can skip them for now, but in the future we can use shell scripting to create the required package from `pybamm-cookiecutter` and check if it is installable. Similarly for licenses, we can check if the created package has the selected license.

PR #7, opened 2023-08-14
feat: VCS versioning and mypy

Saransh-cpp:
VCS versioning by default for the project and the template.

mypy for the project (I think adding an option for the template would make this PR huge, but I'll add it through another PR 🙂)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -126,3 +126,6 @@ cython_debug/
 
 # Code editor files and folders
 .vscode/
+
+# hatch-vcs
+src/pybamm_cookiecutter/_version.py
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,6 +37,7 @@ repos:
     rev: v0.0.281
     hooks:
       - id: ruff
+        files: "^(src/pybamm_cookiecutter|tests$)"
         args: ["--fix", "--show-fixes"]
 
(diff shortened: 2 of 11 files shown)


agriyakhetarpal (approved):
Wow nice there are a lot of changes. Making `mypy` optional in the template would be great, especially because we were not using it for the project (should we do that?). Also, could you clarify if there will be changes on how we update versions while making a release with the addition of `hatch-vcs`—is there anything to keep in mind after one updates the version in `__init__.py`, because I read it uses timestamps from version control systems? That way I will include relevant resources when I prepare the documentation for using the template. Thanks for doing this, @Saransh-cpp!

Saransh-cpp:
> especially because we were not using it for the project (should we do that?)

This PR adds mypy for the project!

> Also, could you clarify if there will be changes on how we update versions while making a release with the addition of hatch-vcs—is there anything to keep in mind after one updates the version in __init__.py

Create a new release on GitHub and done 🙂 

`hatch-vcs` will automatically fetch that git tag and bump the version. Or, if you are working locally, create a new `git tag` and that is your new version!

agriyakhetarpal:
That sounds great, will simplify a lot of things. The changes look good to me, feel free to merge whenever

Saransh-cpp:
Thanks for the review! Merging.

PR #8, opened 2023-09-05
feat: enable mypy for the template

Saransh-cpp:
Following up on #7

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -17,7 +17,7 @@
     "backend": [
         "hatch"
     ],
-    "mypy": true,
+    "mypy": false,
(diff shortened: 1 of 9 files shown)


cookiecutter.json
```diff
@@ -17,6 +17,7 @@
     "backend": [
         "hatch"
     ],
+    "mypy": true,
```
agriyakhetarpal, line 20: Could this be made false by default, perhaps?

agriyakhetarpal (approved):
Thanks, @Saransh-cpp! I got to this a bit late, apologies. Looks good to me overall; I added a minor comment, feel free to merge after addressing it 😄

PR #12, opened 2024-06-13
Added workflows

santacodes:
The workflow wasn't working on the main branch because I missed the workflows directory under `.github` while switching between branches. Hope it works now.


agriyakhetarpal:
Workflow triggered on `main`: https://github.com/pybamm-team/pybamm-cookiecutter/actions/runs/9503637963/job/26194513060

PR #13, opened 2024-06-13
Fixed style issues

santacodes:
- Fixed style issues in the last commit of `main` branch


santacodes:
It seems the Linux and Windows tests are failing because `uv` can't work with the cache files of `pip`. I tried accessing `uv`'s cache on my fork but it doesn't seem to work as well. So I guess it is better to remove caching completely until there is an update on `uv`'s caching officially.

agriyakhetarpal (approved):

PR #15, opened 2024-06-20
Added parameter entry points

santacodes:
Additions - 
+ `parameter_sets.py` from the main PyBaMM repository which acts as an entry point for the parameter sets mapped through the group `cookie_parameter_sets`.
+ `Chen2020.py` to test and load a parameter set through the entry point to ensure that everything works. 
+ Tests to test the entry points.

Updates - 
- A nox `dev` session. 
- Updated rules in compliance with the main PyBaMM repository in `pyproject.toml`.

Accessing parameter sets through entry points here is the same as accessing it through PyBaMM. Instead using `pybamm.parameter_sets` here it would imply `pybamm_cookiecutter.paramter_sets` to access them. I think it would make it easier for current users to adapt to the template as the implementations are almost the same. Before release, the template should be modified in compliance with the module name a user would give.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2023-2024, PyBaMM Team
+Copyright (c) 2024, PyBaMM Team
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2023, PyBaMM Team
+Copyright (c) 2023-2024, PyBaMM Team
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
--- a/LICENSES-bundled.txt
+++ b/LICENSES-bundled.txt
@@ -1,0 +1,7 @@
+This project and source distributions bundle several libraries that are
+compatibly licensed.
+
+
+Name: PyBaMM
+Files: src/pybamm_cookiecutter/parameters/*
+License: BSD-3-Clause
--- a/noxfile.py
+++ b/noxfile.py
@@ -37,4 +37,4 @@ def set_dev(session):
     session.install("virtualenv")
     session.run("virtualenv", os.fsdecode(VENV_DIR), silent=True)
     python = os.fsdecode(VENV_DIR.joinpath("bin/python"))
-    session.run(python, "-m", "pip", "install", "-e", ".[dev]")
+    session.run(python, "-m", "uv", "pip", "install", "-e", ".[dev]")
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,8 +1,12 @@
 import nox
+from pathlib import Path
+import os
 
 # Options to modify nox behaviour
 nox.options.default_venv_backend = "uv|virtualenv"
 nox.options.reuse_existing_virtualenvs = True
+
+VENV_DIR = Path("./venv").resolve()
 
 @nox.session(name="docs")
 def build_docs(session: nox.Session) -> None:
@@ -26,3 +30,11 @@ def run_template_generation(session):
     session.install("setuptools", silent=False)
     session.install("-e", ".[dev]", silent=False)
     session.run("pytest", "tests")
+
+@nox.session(name="dev")
+def set_dev(session):
+    """Install pybamm-cookiecutter in editable mode"""
+    session.install("virtualenv")
+    session.run("virtualenv", os.fsdecode(VENV_DIR), silent=True)
+    python = os.fsdecode(VENV_DIR.joinpath("bin/python"))
+    session.run(python, "-m", "pip", "install", "-e", ".[dev]")
--- a/noxfile.py
+++ b/noxfile.py
@@ -37,4 +37,4 @@ def set_dev(session):
     session.install("virtualenv")
     session.run("virtualenv", os.fsdecode(VENV_DIR), silent=True)
     python = os.fsdecode(VENV_DIR.joinpath("bin/python"))
-    session.run(python, "-m", "uv", "pip", "install", "-e", ".[dev]")
+    session.run(python, "-m", "pip", "install", "-e", ".[dev]")
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,20 +109,7 @@ flake8-unused-arguments.ignore-variadic-names = true
 [tool.ruff.lint]
 extend-select = [
     "B",             # flake8-bugbear
-    # "I",           # isort
-    # "ARG",         # flake8-unused-arguments
-    # "C4",          # flake8-comprehensions
-    # "ICN",         # flake8-import-conventions
-    # "ISC",         # flake8-implicit-str-concat
-    # "PGH",         # pygrep-hooks
-    # "PIE",         # flake8-pie
-    # "PL",          # pylint
-    # "PT",          # flake8-pytest-style
-    # "PTH",         # flake8-use-pathlib
-    # "RET",         # flake8-return
     "RUF",           # Ruff-specific
-    # "SIM",         # flake8-simplify
-    # "T20",         # flake8-print
     "UP",            # pyupgrade
     "YTT",           # flake8-2020
     "TID252",        # relative-imports
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,6 +61,9 @@ Homepage = "https://github.com/pybamm-team/pybamm-cookiecutter"
 Discussions = "https://github.com/pybamm-team/pybamm-cookiecutter/discussions"
 Changelog = "https://github.com/pybamm-team/pybamm-cookiecutter/releases"
 
+[project.entry-points."cookie_parameter_sets"]
+Chen2020 = "pybamm_cookiecutter.parameters.input.Chen2020:get_parameter_values"
+
 [tool.hatch]
 version.source = "vcs"
 build.hooks.vcs.version-file = "src/pybamm_cookiecutter/_version.py"
@@ -73,8 +76,8 @@ packages = [
   "src/pybamm_cookiecutter",
   "tests"
 ]
-python_version = "3.8"
-strict = true
+python_version = "3.11"
+strict = false
 warn_return_any = false
 show_error_codes = true
 enable_error_code = [
@@ -84,6 +87,9 @@ enable_error_code = [
 ]
 disallow_untyped_defs = false
 disallow_untyped_calls = false
+ignore_missing_imports = true
+allow_redefinition = true
+disable_error_code = ["call-overload", "operator"]
 
 [tool.coverage]
 run.source = ["pybamm_cookiecutter"]
@@ -92,32 +98,53 @@ port.exclude_lines = [
 ]
 
 [tool.ruff]
-select = [
-  "E", "F", "W", # flake8
-  "B",           # flake8-bugbear
-  "I",           # isort
-  "ARG",         # flake8-unused-arguments
-  "C4",          # flake8-comprehensions
-  "EM",          # flake8-errmsg
-  "ICN",         # flake8-import-conventions
-  "ISC",         # flake8-implicit-str-concat
-  "G",           # flake8-logging-format
-  "PGH",         # pygrep-hooks
-  "PIE",         # flake8-pie
-  "PL",          # pylint
-  "PT",          # flake8-pytest-style
-  "PTH",         # flake8-use-pathlib
-  "RET",         # flake8-return
-  "RUF",         # Ruff-specific
-  "SIM",         # flake8-simplify
-  "T20",         # flake8-print
-  "UP",          # pyupgrade
-  "YTT",         # flake8-2020
-  "EXE",         # flake8-executable
-  "NPY",         # NumPy specific rules
-  "PD",          # pandas-vet
-]
+extend-include = ["*.ipynb"]
+extend-exclude = ["__init__.py"]
 src = ["src"]
 exclude = []
 isort.required-imports = ["from __future__ import annotations"]
 flake8-unused-arguments.ignore-variadic-names = true
+
+
+[tool.ruff.lint]
+extend-select = [
+    "B",             # flake8-bugbear
+    # "I",           # isort
+    # "ARG",         # flake8-unused-arguments
+    # "C4",          # flake8-comprehensions
+    # "ICN",         # flake8-import-conventions
+    # "ISC",         # flake8-implicit-str-concat
+    # "PGH",         # pygrep-hooks
+    # "PIE",         # flake8-pie
+    # "PL",          # pylint
+    # "PT",          # flake8-pytest-style
+    # "PTH",         # flake8-use-pathlib
+    # "RET",         # flake8-return
+    "RUF",           # Ruff-specific
+    # "SIM",         # flake8-simplify
+    # "T20",         # flake8-print
+    "UP",            # pyupgrade
+    "YTT",           # flake8-2020
+    "TID252",        # relative-imports
+]
+ignore = [
+    "E741",        # Ambiguous variable name
+    "RUF012",      # Mutable class attributes should be annotated with `typing.ClassVar`
+    "SIM108",      # Use ternary operator
+    "ARG001",      # Unused function argument:
+    "ARG002",      # Unused method arguments
+    "PLR2004",     # Magic value used in comparison
+    "PLR0915",     # Too many statements
+    "PLR0913",     # Too many arguments
+    "PLR0912",     # Too many branches
+    "RET504",      # Unnecessary assignment
+    "RET505",      # Unnecessary `else`
+    "RET506",      # Unnecessary `elif`
+    "B018",        # Found useless expression
+    "RUF002",      # Docstring contains ambiguous
+    "UP007",       # For pyupgrade
+]
+
+[tool.ruff.lint.per-file-ignores]
+"tests/*" = ["T20"]
+"docs/*" = ["T20"]
--- a/src/pybamm_cookiecutter/parameters/__init__.py
+++ b/src/pybamm_cookiecutter/parameters/__init__.py
@@ -1,0 +1,3 @@
+from __future__ import annotations
+
+__all__ = ["parameter_sets", ]
--- a/src/pybamm_cookiecutter/parameters/__init__.py
+++ b/src/pybamm_cookiecutter/parameters/__init__.py
@@ -1,3 +1,3 @@
 from __future__ import annotations
 
-__all__ = ["parameter_sets", ]
+__all__ = ["parameter_sets",]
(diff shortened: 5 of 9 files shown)


santacodes:
@arjxn-py @brosaplanella could you please review this PR?

src/pybamm_cookiecutter/parameters/parameter_sets.py
agriyakhetarpal, line 1: Since the code for this file is derived that in PyBaMM, we shall need to PyBaMM's BSD-3-Clause license and provide attribution. Could you add a comment about this, add the license based on the template at https://opensource.org/license/bsd-3-clause, and add a `LICENSES-bundled.txt` file in the project root with: the path to the file, the SPDX license identifier (BSD-3-Clause in this case), and the name of the project (PyBaMM)?

This license must be included in our distribution for this project in source code (sdist) and binary form (wheels) respectively. You might need to browse the documentation in https://hatch.pypa.io/latest/plugins/builder/sdist/#default-file-selection and build these distribution formats (`pip install build` and `python -m build --sdist --wheel` to ensure that `LICENSES-bundled` is present.

agriyakhetarpal: and similarly for `Chen2020.py` (though I am not sure right now about what the associated paper at https://iopscience.iop.org/article/10.1149/1945-7111/ab9050/pdf/ about the experimental data)

src/pybamm_cookiecutter/parameters/parameter_sets.py
```diff
@@ -0,0 +1,101 @@
...
+            # this file now return the name as a string, which will load the same
+            # parameter set as before when passed to `ParameterValues`
+            if name in self:
+                msg = (
+                    f"Parameter sets should be called directly by their name ({name}), "
+                    f"instead of via pybamm_cookiecutter.parameter_sets (pybamm_cookiecutter.parameter_sets.{name})."
+                )
+                warnings.warn(msg, DeprecationWarning, stacklevel=2)
+                return name
+            raise error
```
agriyakhetarpal, line 98: Since we will be supporting recent versions of PyBaMM – say, `>=23.5`, perhaps we can remove this check for `pybamm-cookiecutter`'s case here? @brosaplanella

tests/test_entry_points.py
```diff
@@ -0,0 +1,27 @@
...
+import glob
+import importlib.util
+import sys
+
+def test_entry_points():
+    """Testing if the entry points are loading right"""
+
+    entry_points = list(pybamm_cookiecutter.parameter_sets)
+    parameter_sets = glob.glob("src/pybamm_cookiecutter/parameters/input/*.py")
+    # Making a list Parameter sets in the parameters/input directory
```
agriyakhetarpal, line 12: Might be easier to pass a `pathlib.Path` here or use `Path.glob()`

pyproject.toml
```diff
@@ -92,32 +98,53 @@ port.exclude_lines = [
...
+    # "PL",          # pylint
+    # "PT",          # flake8-pytest-style
+    # "PTH",         # flake8-use-pathlib
+    # "RET",         # flake8-return
+    "RUF",           # Ruff-specific
+    # "SIM",         # flake8-simplify
+    # "T20",         # flake8-print
+    "UP",            # pyupgrade
+    "YTT",           # flake8-2020
+    "TID252",        # relative-imports
```
brosaplanella, line 128: Do we need to keep the comment ones or can we delete them?

src/pybamm_cookiecutter/parameters/parameter_sets.py
brosaplanella: You mean to give attribution for the parameter set?

src/pybamm_cookiecutter/parameters/parameter_sets.py
brosaplanella: Given that the cookiecutter is under development, I think it is fine to only support the most recent version(s) for now.

brosaplanella (commented):
Nice work, added some additional comments to Agriya's.

src/pybamm_cookiecutter/parameters/parameter_sets.py
agriyakhetarpal: Yes. The paper says that the work is distributed under CC BY-NC-ND, but the code that PyBaMM uses to provide the work as a parameter set is under BSD, hence the non-clarity

src/pybamm_cookiecutter/parameters/parameter_sets.py
```diff
@@ -0,0 +1,120 @@
...
+    def __new__(cls):
+        """Ensure only one instance of ParameterSets exists"""
+        if not hasattr(cls, "instance"):
+            cls.instance = super().__new__(cls)
+        return cls.instance
+
+    def __getitem__(self, key) -> dict:
+        return self.__load_entry_point__(key)()
+
+    def __load_entry_point__(self, key) -> Callable:
```
arjxn-py, line 90:
```suggestion
    def _load_entry_point(self, key) -> Callable:
```
double underscores are not typical for non-magic methods. I'd suggest it renaming as a private method instead.

src/pybamm_cookiecutter/parameters/parameter_sets.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def get_docstring(self, key):
+        """Return the docstring for the ``key`` parameter set"""
+        return textwrap.dedent(self.__load_entry_point__(key).__doc__)
+
+    def __getattribute__(self, name):
+        try:
+            return super().__getattribute__(name)
+        except AttributeError as error:
+              raise error
```
arjxn-py, line 117:
```suggestion
    def __getattribute__(self, name: str) -> Any:
        try:
            return super().__getattribute__(name)
        except AttributeError as error:
            if name in self:
                msg = (
                    f"Parameter sets can be called directly by their name ({name}), "
                    f"instead of via pybamm_cookiecutter.parameter_sets (pybamm_cookiecutter.parameter_sets.{name})."
                )
                warnings.warn(msg, DeprecationWarning, stacklevel=2)
                return name
            raise error
```
Does this look better? Make sure to test it with edge cases too.

src/pybamm_cookiecutter/parameters/parameter_sets.py
```diff
@@ -0,0 +1,120 @@
...
+        Parameters for the Enertech cell (Ai2020), from the papers :footcite:t:`Ai2019`,
+        :footcite:t:`rieger2016new` and references therein.
+        ...
+
+    See also: :ref:`adding-parameter-sets`
+
+    """
+
+    def __init__(self):
+        # Dict of entry points for parameter sets, lazily load entry points as
```
arjxn-py, line 68:
```suggestion
        '''Dict of entry points for parameter sets, lazily load entry points as'''
```
Is this docstring complete?

src/pybamm_cookiecutter/parameters/parameter_sets.py
```diff
@@ -0,0 +1,120 @@
...
+        self.__all_parameter_sets = dict()
+        for entry_point in self.get_entries("cookie_parameter_sets"):
+            self.__all_parameter_sets[entry_point.name] = entry_point
+
+    @staticmethod
+    def get_entries(group_name):
+        # Wrapper for the importlib version logic
```
arjxn-py, line 75:
```suggestion
        '''Wrapper for the importlib version logic'''
```

src/pybamm_cookiecutter/parameters/parameter_sets.py
```diff
@@ -0,0 +1,120 @@
...
+        if sys.version_info < (3, 10):  # pragma: no cover
```
arjxn-py, line 76: You can add testing for this line by maybe manipulating `sys.version_info` in tests?

tests/test_entry_points.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+import pybamm_cookiecutter
+import importlib.util
+import sys
+from pathlib import Path
+
+def test_entry_points():
+    """Testing if the entry points are loading right"""
```
arjxn-py, line 8:
```suggestion
    """Test if the entry points are loaded correctly."""
```
Nitpicking 😅

noxfile.py
```diff
@@ -1,9 +1,13 @@
...
+from pathlib import Path
+import os
 
 # Options to modify nox behaviour
 nox.options.default_venv_backend = "uv|virtualenv"
 nox.options.reuse_existing_virtualenvs = True
 
+VENV_DIR = Path("./venv").resolve()
+
 @nox.session(name="docs")
```
arjxn-py, line 11:
```suggestion
@nox.session(name="docs", reuse_venv=True)
```

noxfile.py
```diff
@@ -26,3 +30,11 @@ def run_template_generation(session):
     session.install("setuptools", silent=False)
     session.install("-e", ".[dev]", silent=False)
     session.run("pytest", "tests")
+
+@nox.session(name="dev")
```
arjxn-py, line 34:
```suggestion
@nox.session(name="dev", reuse_venv=True)
```
I think it should be good to have `reuse_venv=True` in sessions so feel free to do that for others too

agriyakhetarpal: We already have `nox.options.reuse_existing_virtualenvs = True` at the top as a global setting so this argument would not be needed

src/pybamm_cookiecutter/parameters/parameter_sets.py
santacodes: Ferran and Agriya wanted to follow the latest PyBaMM release for this project hence I removed that check. You can refer to that discussion [here](https://github.com/pybamm-team/pybamm-cookiecutter/pull/15#discussion_r1649772356).

src/pybamm_cookiecutter/parameters/parameter_sets.py
agriyakhetarpal: I don't have enough access to configure it, so I assume you wouldn't either. @brosaplanella, could you help us configure CodeCov? If it helps, here's the link to it: https://github.com/pybamm-team/pybamm-cookiecutter/settings/installations

src/pybamm_cookiecutter/parameters/parameter_sets.py
arjxn-py: Didn't saw the other discussion, hence marking this one as resolved.

LICENSES-bundled.txt
```diff
@@ -0,0 +1,7 @@
+This project and source distributions bundle several libraries that are
+compatibly licensed.
+
+
+Name: PyBaMM
```
brosaplanella, line 5: Given that PyBaMM will install some dependencies, do we need to list them here too?

brosaplanella (approved):
Just a small comment, otherwise happy to merge if Agriya and Arjun are happy with it too.

LICENSES-bundled.txt
agriyakhetarpal: I don't think so, because we don't use the source code from those dependencies – it is just PyBaMM.

agriyakhetarpal: I don't think so, because we don't use the source code from those dependencies – it is just PyBaMM.

LICENSE
```diff
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2023, PyBaMM Team
+Copyright (c) 2023-2024, PyBaMM Team
```
agriyakhetarpal, line 3:
```suggestion
Copyright (c) 2024, PyBaMM Team
```

noxfile.py
```diff
@@ -26,3 +30,11 @@ def run_template_generation(session):
...
+def set_dev(session):
+    """Install pybamm-cookiecutter in editable mode"""
+    session.install("virtualenv")
+    session.run("virtualenv", os.fsdecode(VENV_DIR), silent=True)
+    python = os.fsdecode(VENV_DIR.joinpath("bin/python"))
+    session.run(python, "-m", "uv", "pip", "install", "-e", ".[dev]")
```
agriyakhetarpal, line 40: I tried this locally and it isn't working – `uv` doesn't seem to properly install dependencies into a `virtualenv`-created environment. I think it needs itself to create the virtual environment, i.e., using `uv venv`. I don't wish to hold the PR, so maybe we can use the original, non-`uv` solution and fix this later?

src/pybamm_cookiecutter/parameters/__init__.py
```diff
@@ -0,0 +1,3 @@
+from __future__ import annotations
+
+__all__ = ["parameter_sets", ]
```
agriyakhetarpal, line 3:
```suggestion
__all__ = ["parameter_sets",]
```

(nit)

src/pybamm_cookiecutter/parameters/input/Chen2020.py
```diff
@@ -0,0 +1,369 @@
+"""
+This code is adopted from the PyBaMM project under the BSD 3-Clause
```
agriyakhetarpal, line 2:
```suggestion
This code is adopted from the PyBaMM project under the BSD-3-Clause
```

The correct SPDX short identifier is "BSD-3-Clause", so either this should be "BSD-3-Clause" as per my suggestion or the entire name, 'BSD 3-Clause "New" or "Revised" License'.

See https://spdx.org/licenses/BSD-3-Clause.html

src/pybamm_cookiecutter/parameters/parameter_sets.py
```diff
@@ -0,0 +1,118 @@
+"""
+This code is adopted from the PyBaMM project under the BSD 3-Clause
```
agriyakhetarpal, line 2: Same here

src/pybamm_cookiecutter/parameters/parameter_sets.py
```diff
@@ -0,0 +1,118 @@
...
+        Parameters for the Enertech cell (Ai2020), from the papers :footcite:t:`Ai2019`,
+        :footcite:t:`rieger2016new` and references therein.
+        ...
+
+    See also: :ref:`adding-parameter-sets`
+
+    """
+
+    def __init__(self):
+        '''Dict of entry points for parameter sets, lazily load entry points as'''
```
agriyakhetarpal, line 66:
```suggestion
        """Dict of entry points for parameter sets, lazily load entry points as"""
```

I would suggest using double quotes instead of single-quotes to keep consistency with the docstring above this function.

src/pybamm_cookiecutter/parameters/parameter_sets.py
```diff
@@ -0,0 +1,118 @@
...
+        self.__all_parameter_sets = dict()
+        for entry_point in self.get_entries("cookie_parameter_sets"):
+            self.__all_parameter_sets[entry_point.name] = entry_point
+
+    @staticmethod
+    def get_entries(group_name):
+        '''Wrapper for the importlib version logic'''
```
agriyakhetarpal, line 73:
```suggestion
        """Wrapper for the importlib version logic"""
```

src/pybamm_cookiecutter/parameters/parameter_sets.py
```diff
@@ -0,0 +1,118 @@
...
+        >>> list(pybamm_cookiecutter.parameter_sets)
+        ['Chen2020', ...]
+
+    Get the docstring for a parameter set:
+
+
+        >>> print(pybamm_cookiecutter.parameter_sets.get_docstring("Ai2020"))
+        <BLANKLINE>
+        Parameters for the Enertech cell (Ai2020), from the papers :footcite:t:`Ai2019`,
+        :footcite:t:`rieger2016new` and references therein.
```
agriyakhetarpal, line 58: We don't have doctests or a citations file set up. I think we should set them up after the structure is ready. Let's keep these in mind

src/pybamm_cookiecutter/parameters/parameter_sets.py
```diff
@@ -0,0 +1,118 @@
...
+            cls.instance = super().__new__(cls)
+        return cls.instance
+
+    def __getitem__(self, key) -> dict:
+        return self._load_entry_point(key)()
+
+    def _load_entry_point(self, key) -> Callable:
+        """Check that ``key`` is a registered ``cookie_parameter_sets``,
+        and return the entry point for the parameter set, loading it needed.
+        """
```
agriyakhetarpal, line 91:
```suggestion
        """Check that ``key`` is a registered ``cookie_parameter_sets``,
        and return the entry point for the parameter set, loading it needed."""
```

(nit) move the ending quote to the same line for consistency

tests/test_entry_points.py
```diff
@@ -0,0 +1,27 @@
...
+import importlib.util
+import sys
+from pathlib import Path
+
+def test_entry_points():
+    """Test if the entry points are loaded correctly."""
+
+    entry_points = list(pybamm_cookiecutter.parameter_sets)
+    parameter_sets = Path("src/pybamm_cookiecutter/parameters/input/").glob("*.py")
+    # Making a list Parameter sets in the parameters/input directory
```
agriyakhetarpal, line 12:
```suggestion
    # Making a list of parameter sets in the parameters/input directory
```

tests/test_entry_points.py
```diff
@@ -0,0 +1,27 @@
...
+    """Testing if the values get loaded via parameter entry points and are equal when loaded through entry points"""
+    # Loading parameter_sets through entry points
+    parameters = pybamm_cookiecutter.parameter_sets['Chen2020']
+    # Loading parameter sets through the source file by dynamically loading Chen2020.py as a module
+    spec = importlib.util.spec_from_file_location("Chen2020mod", "src/pybamm_cookiecutter/parameters/input/Chen2020.py")
+    chen_module = importlib.util.module_from_spec(spec)
+    sys.modules["Chen2020mod"] = chen_module
+    spec.loader.exec_module(chen_module)
+    parameters_from_file = chen_module.get_parameter_values()
+    assert parameters.keys() == parameters_from_file.keys(), "The keys in the module and local input file are not the same"
```
agriyakhetarpal, line 27: Could we improve this error message? We can add something like, "Got {}, expected {}". Not sure if it is needed with `pytest`, though.

agriyakhetarpal (commented):
Looks good to me too, a few final comments! Thanks, @santacodes!

noxfile.py
santacodes: Strange, it works for me when I use nox outside the `virtualenv` and it detects the `venv` folder in the working directory and installs the project inside it, perhaps we should look at this later when we are done with model entry points.

agriyakhetarpal (approved):

src/pybamm_cookiecutter/parameters/parameter_sets.py
arjxn-py: Resolving this as being tracked in #19

src/pybamm_cookiecutter/parameters/parameter_sets.py
arjxn-py: This should be covered in #20

noxfile.py
arjxn-py: I don't want to hold this PR either, but would it be fine to test this session on CI as well?

arjxn-py: Resolving this, at the same time RFD on #21

arjxn-py (approved):
Thanks @santacodes, merging this as you must have also started with model entry points so didn't want to hold this.
I hope it didn't raise any bottlenecks.

PR #19, opened 2024-07-03
Added CodeCov workflow 

santacodes:
+ Set up codecov workflow as discussed in today's meet
+ Added a `codecov.yml` config file

### DIFF
--- a/.github/workflows/test_on_push.yml
+++ b/.github/workflows/test_on_push.yml
@@ -54,6 +54,10 @@ jobs:
         run: |
           nox -s test-generation
 
+      - name: Run coverage tests
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12'
+        run: python -m nox -s coverage
+
       - name: Upload coverage report
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12'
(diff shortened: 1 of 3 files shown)


.github/workflows/test_on_push.yml
```diff
@@ -52,3 +53,9 @@ jobs:
       - name: Test Template Generation
         run: |
           nox -s test-generation
+
+      - name: Upload coverage report
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12'
+        uses: codecov/codecov-action@v4.5.0
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
agriyakhetarpal, line 61: Are we using the `coverage` module so far? I think we need to add it and integrate it with the tests first for any meaningful report.

agriyakhetarpal: Since the workflow and the GitHub app have been added which brought the primary purpose, I think we should merge this first and see how it works for any of the other PRs, and then fix things as necessary.

agriyakhetarpal (approved):

arjxn-py (approved):
Sure, happy to merge this. Thanks @santacodes

PR #23, opened 2024-07-09
Added Model Entry Points

santacodes:
+ Implemented `Model Entry Points` to create Model objects within the template and initialise them through entry points.
+ Added the `SPM` model which can be initialised through the model entry points.
+ A wrapper method to load a model object called `models("modelname/authorname")` is added.
Example - 
To load the `SPM` model, after installing the `pybamm_cookiecutter` project, it can be accessed by calling, `pybamm_cookiecutter.Model("SPM")`. This would return an initialised model object of the `SPM` model.
+ Added two basic tests for model entry points.


valentinsulzer:
Nice to see this progressing! Good to prototype this in `pybamm-cookiecutter` but once it works well we should move the entry point logic to PyBaMM itself to match the `ParameterValues` entry points, for example users should be able to do `pybamm.Model("author/name")`

santacodes:
Tweaked some things to unify both the entry points for `parameter_sets` and `models`, I noticed `pybamm.parameter_sets` is a `singleton instance` for loading `parameter_sets`, so I made some changes to make it a `doubleton instance` to accommodate `Model Entry Points` and `Parameter Sets Entry Points` in the same file, as in through the same entry point, just two different instances. I tested the logic and it seems to be working, if you create a third instance through the entry point it will just return you the object of the second instance(the most recent instance that was created).
third_instance

You can load a `Model` by calling the `pybamm_cookiecutter.Model('SPM')` method, and load a `parameter_set` by doing `pybamm_cookiecutter.parameter_sets["Chen2020"]` the same way we do it in `PyBaMM`.

src/pybamm_cookiecutter/__init__.py
```diff
@@ -8,10 +8,12 @@
...
 
 from ._version import version as __version__
-from .parameters.parameter_sets import parameter_sets
+from .entry_point import Model, parameter_sets, models
 
 __all__ : list[str] = [
     "__version__",
     "pybamm",
     "parameter_sets",
+    "Model",
```
brosaplanella, line 17: Why do we have `"Model"` and `"models"`? Following my previous comment, I would rather stick to the latter.

src/pybamm_cookiecutter/entry_point.py
```diff
@@ -76,35 +84,35 @@ def get_entries(group_name):
         else:
             return importlib.metadata.entry_points(group=group_name)
 
-    def __new__(cls):
-        """Ensure only one instance of ParameterSets exists"""
-        if not hasattr(cls, "instance"):
+    def __new__(cls, group):
+        """Ensure only two instances of entry points exist, one for parameter sets and the other for models"""
+        if EntryPoint._instances < 2:
```
brosaplanella, line 89: Given that is now <2, is there any chance that someone accidentally initialises 2 instances of parameter sets? As in, would it be safer to check that each can only be initialised once?

src/pybamm_cookiecutter/entry_point.py
```diff
@@ -37,19 +37,22 @@
...
     --------
     Listing available parameter sets:
         >>> import pybamm_cookiecutter
         >>> list(pybamm_cookiecutter.parameter_sets)
         ['Chen2020', ...]
+        >>> list(pybamm_cookiecutter.models)
+        ['SPM', ...]
 
-    Get the docstring for a parameter set:
+    Get the docstring for a parameter set/model:
```
agriyakhetarpal, line 55: This line mentions that one can get a docstring for both a parameter set or a model, so we should provide an additional example similar to the one below.

src/pybamm_cookiecutter/entry_point.py
agriyakhetarpal: I think one can add multiple parameter sets (or models) inside one entry point group, so this code is trying to check if someone defines the `[project.entry-points."parameter_sets"]` (or the models counterpart) table more than once in `pyproject.toml`. Is that correct, @santacodes?

src/pybamm_cookiecutter/entry_point.py
```diff
@@ -114,4 +122,22 @@ def __getattribute__(self, name):
...
+    Returns the loaded model object
+
+    Parameters
+    ----------
+    model : str
+        The model name or author name of the model mentioned at the model entry point.
+    Returns
+    -------
+    pybamm.model
+        Model object of the initialised model.
```
agriyakhetarpal, line 141: We should add an "Examples" section here as well so that people know what the format for the model string is (e.g., "Author/Year" or whichever one we decide to use).

agriyakhetarpal (commented):
Slightly nitpicky, but since the API is being merged into one, could we test the model entry points and the parameter sets entry points in the same file?

src/pybamm_cookiecutter/entry_point.py
santacodes: > Given that is now <2, is there any chance that someone accidentally initialises 2 instances of parameter sets? As in, would it be safer to check that each can only be initialised once?

A user cannot initialise 2 instances of parameter sets accidentally through the imported module itself but they definitely can initialise an instance in `entry_point.py`. In that case, only two instances will be accounted and created, the first two ones mentioned in the file, the third one would simply be the same object of the second one initialised. So any new instances other than the first two will just have the same object address of the second instance like depicted in the screenshot. I'm not sure this explanation was clear enough though so please feel free to mention if I said anything wrong 😅.

src/pybamm_cookiecutter/__init__.py
santacodes: `Model()` is a wrapper method that returns the mentioned `Model object` like discussed [here](https://github.com/pybamm-team/pybamm-cookiecutter/pull/23#issuecomment-2218443815).

santacodes: `models` is the model entry points instance that acts like `parameter_sets` instance, so you could get a list of all the listed models in `pyproject.toml` by doing `list(pybamm_cookiecutter.models)` which would return a list of models for e.g `["SPM",...]`. You can initialise a battery model by doing `pybamm_cookiecutter.models["SPM"]` too which is what the wrapper method(`Model()`) basically does.

tests/test_entry_points.py
```diff
@@ -25,3 +25,19 @@ def test_entry_point_load():
...
+    # Making a list Parameter sets in the parameters/input directory
+    models = [x.stem for x in models]
+
+    assert models == entry_points, "Entry points missing either in pyproject.toml or in the input directory"
+
+def test_model_entry_point_load():
+    """Testing if the model gets initialised and returned."""
+    # Loading parameter_sets through entry points
+    model_instance = pybamm_cookiecutter.Model("SPM")
+    assert model_instance != None
```
agriyakhetarpal, line 43:
```suggestion
    assert model_instance is not None
```

src/pybamm_cookiecutter/entry_point.py
agriyakhetarpal: I think we can do this when we have the docs ready

agriyakhetarpal (approved):

agriyakhetarpal:
I think this looks good to me and we have two approvals too, so I propose that we merge this so that we can start moving these files to the template. It's likely we'll revisit this code many times anyway, so once we are done, we can follow @valentinsulzer's suggestion in https://github.com/pybamm-team/pybamm-cookiecutter/pull/23#issuecomment-2218443815 to add the same things to the PyBaMM repository

PR #25, opened 2024-07-17
Added doctest to CI

### LINKED ISSUE #20: Implement `nox -s docs` in the CI
This issue is just to keep track of the suggestion made in #17

santacodes:
Fixes #20

### DIFF
--- a/.github/workflows/test_on_push.yml
+++ b/.github/workflows/test_on_push.yml
@@ -68,8 +68,6 @@ jobs:
   run_doctests:
     needs: style
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
     name: Doctests (ubuntu-latest / Python 3.12)
 
     steps:
--- a/.github/workflows/test_on_push.yml
+++ b/.github/workflows/test_on_push.yml
@@ -7,6 +7,7 @@ on:
     branches:
     - main
 
+
 jobs:
   style:
     runs-on: ubuntu-latest
@@ -63,3 +64,29 @@ jobs:
         uses: codecov/codecov-action@v4.5.0
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+
+  run_doctests:
+    needs: style
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    name: Doctests (ubuntu-latest / Python 3.11)
+
+    steps:
+      - name: Check out pybamm-cookiecutter repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        id: setup-python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.11
+          cache: 'pip'
+
+      - name: Install nox
+        run: python -m pip install nox
+
+      - name: Check if the documentation can be built for GNU/Linux
+        run: python -m nox -s docs
--- a/.github/workflows/test_on_push.yml
+++ b/.github/workflows/test_on_push.yml
@@ -70,7 +70,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-    name: Doctests (ubuntu-latest / Python 3.11)
+    name: Doctests (ubuntu-latest / Python 3.12)
 
     steps:
       - name: Check out pybamm-cookiecutter repository
@@ -79,14 +79,13 @@ jobs:
           fetch-depth: 0
 
       - name: Set up Python
-        id: setup-python
         uses: actions/setup-python@v5
         with:
-          python-version: 3.11
+          python-version: 3.12
           cache: 'pip'
 
       - name: Install nox
         run: python -m pip install nox
 
-      - name: Check if the documentation can be built for GNU/Linux
+      - name: Check if the documentation can be built
         run: python -m nox -s docs
--- a/README.md
+++ b/README.md
@@ -84,5 +84,6 @@ For any questions, comments, suggestions or bug reports, please see the
 
 ## 📃 License
 
-`pybamm-cookiecutter` is fully open source. For more information about its license, see [LICENSE](https://github.com/pybamm-team/pybamm-cookiecutter/blob/main/LICENSE).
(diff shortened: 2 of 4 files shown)


Saransh-cpp (changes requested):
Thanks, @santacodes! Looks good to me! Could you also add the same to the template?

santacodes:
> Thanks, @santacodes! Looks good to me! Could you also add the same to the template?

Would it be okay to wait for this PR and the model entry points PR to merge? Then we could have a separate PR to migrate from the project to the template with all the workflows, tests, and entry points. In that case, we would just have to copy the contents from the project to the template so we wouldn't have to keep track and hand-pick the things missing in the template from the project. I would also like to hear others' opinions on this too.

.github/workflows/test_on_push.yml
```diff
@@ -63,3 +64,29 @@ jobs:
         uses: codecov/codecov-action@v4.5.0
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+
+  run_doctests:
+    needs: style
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    name: Doctests (ubuntu-latest / Python 3.11)
```
agriyakhetarpal, line 73:
```suggestion
    name: Doctests (ubuntu-latest / Python 3.12)
```

Maybe we can test with Python 3.12 – shouldn't be a problem?

.github/workflows/test_on_push.yml
```diff
@@ -63,3 +64,29 @@ jobs:
...
+      - name: Check out pybamm-cookiecutter repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        id: setup-python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.11
```
agriyakhetarpal, line 85:
```suggestion
          python-version: 3.12
```

README.md
```diff
@@ -82,4 +84,5 @@ For any questions, comments, suggestions or bug reports, please see the
 
 ## 📃 License
 
-`pybamm-cookiecutter` is fully open source. For more information about its license, see [LICENSE](https://github.com/pybamm-team/pybamm-cookiecutter/blob/main/LICENSE.txt).
+`pybamm-cookiecutter` is fully open source. For more information about its license, see [LICENSE](https://github.com/pybamm-team/pybamm-cookiecutter/blob/main/LICENSE).
+<!-- SPHINX-END -->
```
agriyakhetarpal, line 88:
```suggestion

<!-- SPHINX-END -->
```

README.md
```diff
@@ -82,4 +84,5 @@ For any questions, comments, suggestions or bug reports, please see the
 
 ## 📃 License
 
-`pybamm-cookiecutter` is fully open source. For more information about its license, see [LICENSE](https://github.com/pybamm-team/pybamm-cookiecutter/blob/main/LICENSE.txt).
+`pybamm-cookiecutter` is fully open source. For more information about its license, see [LICENSE](https://github.com/pybamm-team/pybamm-cookiecutter/blob/main/LICENSE).
```
agriyakhetarpal, line 87:
```suggestion
The `pybamm-cookiecutter` project is open source code. For more information about its license, see [LICENSE](https://github.com/pybamm-team/pybamm-cookiecutter/blob/main/LICENSE).
```

I think the word "fully" is redundant

noxfile.py
```diff
@@ -13,7 +13,9 @@ def build_docs(session: nox.Session) -> None:
     """Build the documentation and load it in a browser tab, rebuilding on changes."""
     envbindir = session.bin
     session.install("-e", ".[docs]")
-    with session.chdir("docs/"):
+    session.chdir("docs/")
```
agriyakhetarpal, line 16: You can use `session.chdir()` as a context manager here: https://nox.thea.codes/en/stable/config.html#nox.sessions.Session.chdir

noxfile.py
```diff
@@ -23,6 +25,17 @@ def build_docs(session: nox.Session) -> None:
...
+    else:
+        session.run(
+            "sphinx-build",
+            "-b",
+            "html",
+            "-W",
+            "--keep-going",
+            ".",
+            f"{envbindir}/../tmp/html",
+        )
```
agriyakhetarpal, line 38: (nit) I know we use this in PyBaMM too, but I usually find it better to find the built files if they are placed in the `docs/build/html` folder itself, and it makes working without `nox` easier because one doesn't need to go into the `.nox/` folders

.github/workflows/test_on_push.yml
```diff
@@ -63,3 +64,29 @@ jobs:
...
+    name: Doctests (ubuntu-latest / Python 3.11)
+
+    steps:
+      - name: Check out pybamm-cookiecutter repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        id: setup-python
```
agriyakhetarpal, line 82:
```suggestion
```

.github/workflows/test_on_push.yml
```diff
@@ -63,3 +64,29 @@ jobs:
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.11
+          cache: 'pip'
+
+      - name: Install nox
+        run: python -m pip install nox
+
+      - name: Check if the documentation can be built for GNU/Linux
```
agriyakhetarpal, line 91:
```suggestion
      - name: Check if the documentation can be built
```

agriyakhetarpal (approved):

.github/workflows/test_on_push.yml
```diff
@@ -63,3 +64,28 @@ jobs:
         uses: codecov/codecov-action@v4.5.0
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+
+  run_doctests:
+    needs: style
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
agriyakhetarpal, line 72: Not needed since we don't have a matrix

Issue #26, opened 2024-07-18
Tracker for first PyPI release


santacodes:
# Overview
This issue tracks all the tasks that need to be checked before the first PyPI release for the `pybamm-cookie` template.

### Migrations
- [x] Migrate the following contents of the project to the template - 
   - `entry points`, for both `parameter_sets` and `models`. #27 
   - `tests` specifically to be shipped to the user inside the template, i.e. the `entry points` tests and an example test file for the users to add test cases. #29 
   - `models` & `parameter_sets` included in the project, i.e. `SPM` and `Chen2020`. #27 
   - `nox-sessions` that are not present in the template. They would be - `unit tests`(yet to be separated), `coverage` and `dev` session. #29 
   - Move suitable `test_on_push` CI to the template. #29 
 ### Features
- [x] Separate `nox sessions` for **tests for project** and **tests for template**. #29 
- [x] Two steps in the `CI` for **tests for project** and **tests for template**. #29 
- [ ] Add `example notebooks` for using `pybamm-cookie` template and entry points.
- [x] Add the `reservoir model`. #30
- [x] Script for automatic `git init` while project generation using the template. Could be referred [here](https://github.com/osl-incubator/scicookie/blob/428785d48d95fa04d17463dd144b5508ab9c77f5/src/scicookie/hooks/post_gen_project.py#L335-L398). #32 ---> #34
- [x] Add `Dependabot` for CI. #28 
- [x] Migration to `copier`. #34
- [x] CD for publishing. #48 
- [x] Add a wrapper CLI for generating a project using `copier`. #36 
- [x] Add `setuptools` backend. #37 
- [ ] Add a basic CI for `gitlab-ci.yml`.
- [x] Add versioning support for git archives #42

### Misc
- [x] Attribute inspirations #43  #45
- [x] Rename to `pybamm-cookie` #46 
- [x] More documentation in README (maybe the logo too?) #47 

#### PS
If there is something else to add or is missing, please feel free to edit this comment and append it.

santacodes:
@Saransh-cpp @agriyakhetarpal @arjxn-py I looked into `copier` and supporting it would take a fair bit of change in the template.

- Modifying the jinja structure with respect to what `copier` supports, i.e. we would have to change all the variables mostly with the `{{cookiecutter.key}}`. This should be easy though, we don't have to change much, just rename things from `{{ cookiecutter.something }}` -> `{{ something }}`.
[...]

PR #27, opened 2024-07-19
Migrated entry points

santacodes:
Migrated entry points from `project` to the `template`. The generated project only works when the git repository is initialised, so if you are testing it locally do `git init` first before installing it using `pip`.

Subtask of #26

### DIFF
--- a/{{cookiecutter.project_name}}/pyproject.toml
+++ b/{{cookiecutter.project_name}}/pyproject.toml
@@ -119,40 +119,41 @@ port.exclude_lines = [
 ]
 
 [tool.ruff]
-extend-include = ["*.ipynb"]
-extend-exclude = ["__init__.py"]
+select = [
+  "E", "F", "W", # flake8
+  "B",           # flake8-bugbear
+  "I",           # isort
+  #"ARG",         # flake8-unused-arguments
+  "C4",          # flake8-comprehensions
+  "EM",          # flake8-errmsg
+  "ICN",         # flake8-import-conventions
+  "ISC",         # flake8-implicit-str-concat
+  "G",           # flake8-logging-format
+  "PGH",         # pygrep-hooks
+  "PIE",         # flake8-pie
+  "PL",          # pylint
+  "PT",          # flake8-pytest-style
+  "PTH",         # flake8-use-pathlib
+  #"RET",         # flake8-return
+  "RUF",         # Ruff-specific
+  "SIM",         # flake8-simplify
+  "T20",         # flake8-print
+  "UP",          # pyupgrade
+  "YTT",         # flake8-2020
+  "EXE",         # flake8-executable
+  "NPY",         # NumPy specific rules
+  "PD",          # pandas-vet
+]
 src = ["src"]
-exclude = []
-isort.required-imports = ["from __future__ import annotations"]
-flake8-unused-arguments.ignore-variadic-names = true
-
-
-[tool.ruff.lint]
-extend-select = [
-    "B",             # flake8-bugbear
-    "RUF",           # Ruff-specific
-    "UP",            # pyupgrade
-    "YTT",           # flake8-2020
-    "TID252",        # relative-imports
+unfixable = [
+  "T20",  # Removes print statements
+  "F841", # Removes unused variables
 ]
 ignore = [
-    "E741",        # Ambiguous variable name
-    "RUF012",      # Mutable class attributes should be annotated with `typing.ClassVar`
-    "SIM108",      # Use ternary operator
-    "ARG001",      # Unused function argument:
-    "ARG002",      # Unused method arguments
-    "PLR2004",     # Magic value used in comparison
-    "PLR0915",     # Too many statements
-    "PLR0913",     # Too many arguments
-    "PLR0912",     # Too many branches
-    "RET504",      # Unnecessary assignment
-    "RET505",      # Unnecessary `else`
-    "RET506",      # Unnecessary `elif`
-    "B018",        # Found useless expression
-    "RUF002",      # Docstring contains ambiguous
-    "UP007",       # For pyupgrade
+  "E741",        # Ambiguous variable name
+  "E501",        # Line too long
+  "PLR2004",     # Magic value used in comparison
 ]
-
-[tool.ruff.lint.per-file-ignores]
-"tests/*" = ["T20"]
-"docs/*" = ["T20"]
+line-length = 100
+exclude = []
+flake8-unused-arguments.ignore-variadic-names = true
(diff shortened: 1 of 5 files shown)


{{cookiecutter.project_name}}/pyproject.toml
```diff
@@ -108,36 +119,40 @@ port.exclude_lines = [
...
-  "RET",         # flake8-return
-  "RUF",         # Ruff-specific
-  "SIM",         # flake8-simplify
-  "T20",         # flake8-print
-  "UP",          # pyupgrade
-  "YTT",         # flake8-2020
-  "EXE",         # flake8-executable
-  "NPY",         # NumPy specific rules
-  "PD",          # pandas-vet
-]
```
agriyakhetarpal, line 135: Question: should we provide these rules but exclude them by commenting them out?

{{cookiecutter.project_name}}/src/{{cookiecutter.__project_slug}}/models/input/SPM.py
```diff
@@ -0,0 +1,214 @@
...
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+"""
+
+
+#
+# Basic Single Particle Model (SPM)
+#
+import pybamm
+
+class SPM(pybamm.lithium_ion.BaseModel):
```
agriyakhetarpal, line 39:
```suggestion
import pybamm


class SPM(pybamm.lithium_ion.BaseModel):
```

{{cookiecutter.project_name}}/src/{{cookiecutter.__project_slug}}/parameters/input/Chen2020.py
```diff
@@ -0,0 +1,369 @@
...
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+"""
+
+
+from __future__ import annotations
+
+import numpy as np
+import pybamm
+def graphite_LGM50_ocp_Chen2020(sto):
```
agriyakhetarpal, line 38:
```suggestion
import pybamm


def graphite_LGM50_ocp_Chen2020(sto):
```

agriyakhetarpal (approved):

{{cookiecutter.project_name}}/pyproject.toml
Saransh-cpp: I think we should keep all of them un-commented. Is any of these rules causing any trouble in the CI?

{{cookiecutter.project_name}}/pyproject.toml
```diff
@@ -108,36 +119,40 @@ port.exclude_lines = [
...
+    "PLR0915",     # Too many statements
+    "PLR0913",     # Too many arguments
+    "PLR0912",     # Too many branches
+    "RET504",      # Unnecessary assignment
+    "RET505",      # Unnecessary `else`
+    "RET506",      # Unnecessary `elif`
+    "B018",        # Found useless expression
+    "RUF002",      # Docstring contains ambiguous
+    "UP007",       # For pyupgrade
+]
```
Saransh-cpp, line 154: Are all of these causing issues? If not, could you try shrinking down this list to include only the ones that are causing trouble? You can fix the genuine suggestions from ruff in this PR itself.

{{cookiecutter.project_name}}/pyproject.toml
agriyakhetarpal: Some of them were causing trouble because the code is copied from PyBaMM, which does not adhere to these rules. Removing them from the project's `pyproject.toml` is fine, personally. We should provide them in the template so that people can un-comment them at their will if needed.

Saransh-cpp: Oh yes, my bad. I meant the same thing. Keep in the template and remove as per our requirements from the project.

santacodes: The reason I say this is because let us say in the future the entry points are updated upstream in the `PyBaMM` repository and we would want to update that in the template as well but the rules might alter the way it is written so I feel ignoring them would be a better option because most of them would already be linted upstream, rather we could add those files first in `pybamm_cookiecutter` project as it has the same linting rules as `PyBaMM` and then move it inside the template.

Saransh-cpp: I see no harm in doing this; should be okay imo.

{{cookiecutter.project_name}}/src/{{cookiecutter.__project_slug}}/__init__.py
```diff
@@ -8,9 +8,23 @@
...
 {%- else %}
-__all__ = ("__version__",)
+__all__ = [
+    "__version__",
+    "pybamm",
+    "parameter_sets",
+    "Model",
+    "models",
+]
 {%- endif %}
```
Saransh-cpp, line 30: Can this be -
```py
{%- if cookiecutter.mypy %}
__all__: list[str] = [
{%- else %}
__all__ = [
{%- endif %}
    "__version__",
    "pybamm",
    "parameter_sets",
    "Model",
    "models",
]
```

{{cookiecutter.project_name}}/pyproject.toml
```diff
@@ -47,14 +47,15 @@ classifiers = [
 {%- if cookiecutter.backend == "hatch" %}
 dynamic = ["version"]
 {%- endif %}
-dependencies = ["pybamm"]
+dependencies = ["pybamm", "cookiecutter"]
```
Saransh-cpp, line 50: Is the cookiecutter dependency required for the template? Where is it being used?

Saransh-cpp (approved):
Thanks, @santacodes! Looks good to me now! A couple of comments below -

{{cookiecutter.project_name}}/pyproject.toml
santacodes: It is not required imo.

arjxn-py (approved):
Thanks @santacodes, Looks good to me.

{{cookiecutter.project_name}}/pyproject.toml
```diff
@@ -47,14 +47,15 @@ classifiers = [
...
+dependencies = ["pybamm",]
 
 [project.optional-dependencies]
 dev = [
   "pytest >=6",
   "pytest-cov >=3",
-  "nox",
+  "nox[uv]",
   "pre-commit",
+  "pytest-cookies",
```
agriyakhetarpal, line 58:
```suggestion
```

I don't think the generated project will need `pytest-cookies` as a dependency, because the generated project won't have a template to provide to users (and they won't need to test the generation of a project).

agriyakhetarpal (approved):
Looks all good to me, thanks, @santacodes! Just one suggestion on the dependencies for the generated project below

PR #29, opened 2024-07-22
Added new test infrastructure

santacodes:
### Separated `nox sessions` inside `project` and `template`.
+ For `project` : `project-tests` tests the units inside the `pybamm_cookiecutter` project, and `template-tests` tests the `cookiecutter template` for a proper generation.
+ For the `template` : once a user generates a project (lets call it a `generated project`), the `generated project` has two `nox sessions` for the tests, `generated-project-tests` which tests all the units inside the `generated project` and `user-tests` which executes the test cases added by a user for their respective code.
### CI job to test the `generated project`
+ A new CI job to test the `generated project` is added in the workflow, which generates a project using the cookiecutter template and runs the tests contained in the `generated project` to test if the `entry points` and other units are working as intended.
+ CI workflow for the user to test their code has also been added to the template, along with `doctests` and `coverage`.
### Note
The tests running inside `generated project` will fail at the last step as a hook is yet to be configured to initialise `git` inside the template using a `post_gen_project` script, I thought this PR was getting too long to be reviewed so I left that out for a separate one. Also I used `pipx` to run the tests inside the `generated project` because nox wasn't working when changing the working directory to the `generated project`.

Subtask #26


Saransh-cpp:
Nice! I'll get to this once the tests pass.

santacodes:
> Nice! I'll get to this once the tests pass.

The tests are failing because the git initialisation hooks are not added yet, I left it out since this PR was getting too long and was thinking of making a new PR for the hooks, do you want me to add them in this same PR?

Saransh-cpp:
I think it would be better to create and merge a PR for git initialization before this PR goes in/is reviewed.

santacodes:
The tests are now passing after the git initialisation hooks are merged, I will be adding the [patch](https://github.com/pybamm-team/pybamm-cookiecutter/pull/32#issuecomment-2245746685) as I mentioned here to flag out the `hatch-vcs` as an opt-in feature.

.github/workflows/test_on_push.yml
```diff
@@ -51,9 +50,13 @@ jobs:
       - name: Install nox
         run: uv pip install nox[uv]
 
-      - name: Test Template Generation
+      - name: Test template generation
+        run: |
+          nox -s template-tests
+
+      - name: Test project units
```
agriyakhetarpal, line 57:
```suggestion
      - name: Test project
```

.github/workflows/test_on_push.yml
```diff
@@ -87,3 +90,36 @@ jobs:
 
       - name: Check if the documentation can be built
         run: python -m nox -s docs
+
+  generated_project_tests:
+      needs: [style, template_test]
```
agriyakhetarpal, line 95: Since `template_test` needs `style`, you can exclude `style` here and the DAG can then look like this:

```mermaid
flowchart TD;
    A[style] --> B[template_test];
    B --> C[generated_project_tests];
```
instead of this:

```mermaid
flowchart TD;
    A[style] --> C[template_test];
    B[template_test] --> C[generated_project_tests];
```

```suggestion
      needs: [template_test]

```

.github/workflows/test_on_push.yml
```diff
@@ -87,3 +90,36 @@ jobs:
...
+          os: [ubuntu-latest, macos-13, macos-14, windows-latest]
+          python-version: ["3.9", "3.10", "3.11", "3.12"]
+      name:
+        Generated Project Tests (${{ matrix.os }} / Python ${{ matrix.python-version }})
+      steps:
+        - name: Checkout pybamm-cookiecutter
+          uses: actions/checkout@v4
+
+        - name: Set up Python ${{ matrix.python-version }}
+          id: setup-python
```
agriyakhetarpal, line 109:
```suggestion
```

cookiecutter.json
```diff
@@ -1,5 +1,5 @@
 {
-    "project_name": "pybamm-example-project",
+    "project_name": "pybamm_example_project",
```
agriyakhetarpal, line 2: I'm curious, do hyphens in the `project_name` not work? I ask because we do convert the hyphens to underscores in `__project_slug` below.

{{cookiecutter.project_name}}/noxfile.py
```diff
@@ -1,20 +1,67 @@
...
-            "--open-browser",
-            "-qT",
-            ".",
-            f"{envbindir}/../tmp/html",
-        )
+        # For local development
+        if session.interactive:
+            session.run(
+                "sphinx-autobuild",
+                "-j",
```
agriyakhetarpal, line 20: I think we'll either need to list `sphinx-autobuild` as a dependency in `pyproject.toml` or install it here using `session.install("sphinx-autobuild")`.

{{cookiecutter.project_name}}/tests/generated_project_tests/test_entry_points.py
```diff
@@ -0,0 +1,43 @@
+import pytest
+import {{ cookiecutter.project_name }}
+import importlib.util
+import sys
+from pathlib import Path
+
+def test_parameter_sets_entry_points():
```
agriyakhetarpal, line 7:
```suggestion
from pathlib import Path


def test_parameter_sets_entry_points():
```

{{cookiecutter.project_name}}/tests/generated_project_tests/test_entry_points.py
```diff
@@ -0,0 +1,43 @@
...
+    """Test if the parameter_sets via entry points are loaded correctly."""
+
+    entry_points = list({{ cookiecutter.project_name }}.parameter_sets)
+    parameter_sets = Path("src/{{ cookiecutter.project_name }}/parameters/input/").glob("*.py")
+    # Making a list of parameter sets in the parameters/input directory
+    parameter_sets = [x.stem for x in parameter_sets]
+
+    assert parameter_sets == entry_points, "Entry points missing either in pyproject.toml or in the input directory"
+
+def test_parameter_sets_entry_point_load():
```
agriyakhetarpal, line 17:
```suggestion
    assert parameter_sets == entry_points, "Entry points missing either in pyproject.toml or in the input directory"


def test_parameter_sets_entry_point_load():
```

cookiecutter.json
santacodes: Yes I was getting confused between two naming conventions in a lot of different files and modules, `pybamm-example-project` cannot be a module name while `pybamm_example_project` can. So I just made it standard for everything to stick to one naming convention to avoid confusions in multiple files.

{{cookiecutter.project_name}}/noxfile.py
santacodes: It is already a dependency under `[project.optional-dependencies][docs]`, and we are doing `session.install( .. .[docs]` so I think we should be good and it works in CI.

agriyakhetarpal: Never mind, I didn't notice – thanks!

{{cookiecutter.project_name}}/src/{{cookiecutter.__project_slug}}/{% if cookiecutter.backend == "hatch" %}_version.pyi{% endif %}
```diff
@@ -1,4 +1,4 @@
 from __future__ import annotations
 
-version: str
+__version__: str
```
Saransh-cpp, line 3: This file should be generated when `hatch-vcs` is present.
```suggestion
version: str
```

{{cookiecutter.project_name}}/src/{{cookiecutter.__project_slug}}/version.py
```diff
@@ -0,0 +1,6 @@
+from __future__ import annotations
+
+__version__: str
+version_tuple: tuple[int, int, int] | tuple[int, int, int, str, str]
+
+__version__ = '0.1.0'
```
Saransh-cpp, line 6: This file can be deleted.

{{cookiecutter.project_name}}/src/{{cookiecutter.__project_slug}}/__init__.py
```diff
@@ -7,7 +7,7 @@
 from __future__ import annotations
 {%- endif %}
 
-from ._version import version as __version__
+from .version import __version__
```
Saransh-cpp, line 10: Do something like -
```py
if hatch-vcs

from .version import version as __version__

else

__version__: str == "0.0.1"
```

Saransh-cpp (changes requested):

{{cookiecutter.project_name}}/src/{{cookiecutter.__project_slug}}/{% if cookiecutter.hatch_vcs %}_version.py{% endif %}
```diff
@@ -2,3 +2,5 @@ from __future__ import annotations
 
 version: str
 version_tuple: tuple[int, int, int] | tuple[int, int, int, str, str]
+
+version = "0.1.0"
```
Saransh-cpp, line 6: We don't need this, hatch-vcs will take care of the versioning for us.
```suggestion
```

{{cookiecutter.project_name}}/.github/workflows/test_on_push.yml
```diff
@@ -0,0 +1,92 @@
...
+  run_doctests:
+    needs: style
+    runs-on: ubuntu-latest
+    name: Doctests (ubuntu-latest / Python 3.12)
+
+    steps:
+      - name: Check out pybamm-cookiecutter repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
Saransh-cpp, line 80: Don't need the git history for doctests
```suggestion
```

{{cookiecutter.project_name}}/.github/workflows/test_on_push.yml
```diff
@@ -0,0 +1,92 @@
...
+        uses: yezz123/setup-uv@v4
+        with:
+          uv-venv: ".venv"
+
+      - name: Install nox
+        run: uv pip install nox[uv]
+
+      - name: Test project units
+        run: |
+          nox -s generated-project-tests
```
Saransh-cpp, line 55:
```suggestion
        run: nox -s generated-project-tests
```

{{cookiecutter.project_name}}/.github/workflows/test_on_push.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: Test user tests
+        run: |
+          nox -s user-tests
```
Saransh-cpp, line 59:
```suggestion
        run: nox -s user-tests
```

{{cookiecutter.project_name}}/.github/workflows/test_on_push.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: Run coverage tests
+        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12'
+        run: python -m nox -s coverage
```
Saransh-cpp, line 63:
```suggestion
        run: nox -s coverage
```

{{cookiecutter.project_name}}/.github/workflows/test_on_push.yml
```diff
@@ -0,0 +1,92 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.12
+          cache: 'pip'
+
+      - name: Install nox
+        run: python -m pip install nox
+
+      - name: Check if the documentation can be built
+        run: python -m nox -s docs
```
Saransh-cpp, line 92:
```suggestion
        run: nox -s docs
```

{{cookiecutter.project_name}}/noxfile.py
```diff
@@ -1,20 +1,67 @@
...
+    session.install("-e", ".[dev]", silent=False)
+    session.run("pytest", "tests/user_tests")
+
+@nox.session(name="coverage")
+def run_coverage(session):
+    """Run the coverage tests and generate an XML report."""
+    session.install("setuptools", silent=False)
+    session.install("coverage", silent=False)
+    session.install("-e", ".[all,dev,jax]", silent=False)
+    session.run("pytest", "--cov=src/pybamm_cookiecutter", "--cov-report=xml", "tests/")
```
Saransh-cpp, line 59: Cleaner. `coverage` should be installed as a `dev` dependency if it is not being installed in that group at the moment.
```suggestion
    session.posargs.append("--cov=src/pybamm_cookiecutter")
    session.posargs.append("--cov-report=xml")
    run_user_tests(session)
```

{{cookiecutter.project_name}}/.github/workflows/test_on_push.yml
Saransh-cpp: Ditto changes in the other yaml file as well!

Saransh-cpp (changes requested):

{{cookiecutter.project_name}}/.github/workflows/test_on_push.yml
agriyakhetarpal: I think we needed this for the `sphinx-last-updated-by-git` Sphinx extension when I was integrating it last time – I would suggest keeping it for now.

cookiecutter.json
agriyakhetarpal: > pybamm-example-project cannot be a module name while pybamm_example_project can

I think this is a bug, in this case. The name of the project in `pyproject.toml` (and therefore on PyPI) can very well be `pybamm-example-project`, but the files for such a project with hyphens must be placed in `src/pybamm_example_project`, so we are probably not converting the hyphens to underscores (i.e., missing the usage of the the slug variable perhaps).

santacodes: Oops sorry I didn't notice the slug variable was already configured, thanks! I will fix it.

.github/workflows/test_on_push.yml
```diff
@@ -86,4 +87,36 @@ jobs:
         run: python -m pip install nox
 
       - name: Check if the documentation can be built
-        run: python -m nox -s docs
+        run: nox -s docs
+
+  generated_project_tests:
+      needs: [template_test]
```
Saransh-cpp, line 93: Should be independent imo
```suggestion
```

.github/workflows/test_on_push.yml
```diff
@@ -86,4 +87,36 @@ jobs:
...
+          run: python -m pip install cookiecutter pipx
+
+        - name: Generate a project using cookiecutter
+          run: cookiecutter . --no-input
+
+        - name: Install nox and test generated project
+          working-directory: ./pybamm-example-project
+          run: |
+            pipx install nox
+            pipx run nox -s generated-project-tests
```
Saransh-cpp, line 122: We're using `uv` to install `nox` and the nox CLI to execute tests in the job above. Here we're using `pipx` to do our work. Could you make this uniform, please? (use any of the one package downloader/management tool)

cookiecutter.json
```diff
@@ -40,7 +42,8 @@
         "license": "Select a license",
         "backend": {
             "__prompt__": "Choose a build backend",
-            "hatch": "Hatchling (recommended for pure Python projects)"
+            "hatch": "Hatchling (recommended for pure Python projects)",
+              "hatch_vcs": "Enable hatch version control (vcs)?"
```
Saransh-cpp, line 46:
```suggestion
            "hatch": "Hatchling (recommended for pure Python projects)",
            "hatch_vcs": "Enable hatch version control (vcs)?"
```

Saransh-cpp: Is this prompt offered to the users once they select `hatch`? I think you are offering the `hatch-vcs` option under `backend`? Users should select the hatch backend first and then a prompt saying "Enable VCS versioning? (Yes/No)" should enable/disable `hatch-vcs`.

Saransh-cpp: Instead of `hatch_vcs`, just call it `vcs`. This will make it easy to scale with potential new build backends in the future.

Saransh-cpp (changes requested):

.github/workflows/test_on_push.yml
santacodes: I am guessing, if the `template_test` fails, that is if the template itself fails a generation of a project, then there is no purpose of running tests inside the generated project.

cookiecutter.json
santacodes: > Is this prompt offered to the users once they select `hatch`? I think you are offering the `hatch-vcs` option under `backend`? Users should select the hatch backend first and then a prompt saying "Enable VCS versioning? (Yes/No)" should enable/disable `hatch-vcs`.

Yes I was trying to do that and I couldn't find any documentation of nested conditional choices for cookiecutter, though I have gotten it figured out for copier.

Saransh-cpp: Ah no, not necessarily nested, just an independent "vcs" prompt after "hatch" is selected. Treat the build backend and vcs as two independent things, if that makes sense. They should just be prompted to the user in the order 1. backend 2. vcs.

This way, when we add new build backends, say setuptools, the vcs option will update the generated code with setuptools_scm instead of hatch-vcs.

.github/workflows/test_on_push.yml
Saransh-cpp: Oh oh, I see, yes that makes sense. Thanks!

.github/workflows/test_on_push.yml
agriyakhetarpal: +1. Side note: `pipx run nox` means that `pipx install nox` becomes redundant. `pipx run` runs commands from CLI apps without having to install them (though it does install them in a temporary cache, of course, if we are to be pedantic)

cookiecutter.json
```diff
@@ -40,7 +42,8 @@
         "license": "Select a license",
         "backend": {
             "__prompt__": "Choose a build backend",
-            "hatch": "Hatchling (recommended for pure Python projects)"
+            "hatch": "Hatchling (recommended for pure Python projects)",
+              "vcs": "Enable hatch version control (vcs)?"
```
agriyakhetarpal, line 46: I'm not sure if non-experienced packagers will understand what "Enable hatch version control (vcs)?" means. I propose changing it to this:


```suggestion
            "hatch": "Hatchling (recommended for pure Python projects)",
              "vcs": "Let hatch get the version from your version control system (Git/Mercurial/SVN)?"
```

noxfile.py
```diff
@@ -36,12 +36,19 @@ def build_docs(session: nox.Session) -> None:
...
-@nox.session(name="test-generation")
+@nox.session(name="template-tests")
 def run_template_generation(session):
     """Run the tests for testing template generation"""
     session.install("setuptools", silent=False)
     session.install("-e", ".[dev]", silent=False)
-    session.run("pytest", "tests")
+    session.run("pytest", "tests/template_tests")
+
+@nox.session(name="project-tests")
```
agriyakhetarpal, line 46:
```suggestion
    session.run("pytest", "tests/template_tests")


@nox.session(name="project-tests")
```

{{cookiecutter.project_name}}/.github/workflows/test_on_push.yml
```diff
@@ -0,0 +1,90 @@
+name: Project Tests
```
agriyakhetarpal, line 1: This file, too, can benefit from the use of a single package management solution as noted above.

{{cookiecutter.project_name}}/noxfile.py
```diff
@@ -1,20 +1,66 @@
...
+            )
+
+@nox.session(name="generated-project-tests")
+def run_generated_project_tests(session):
+    """Run the tests for testing units inside generated project"""
+    session.install("setuptools", silent=False)
+    session.install("-e", ".[dev]", silent=False)
+    session.run("pytest", "tests/generated_project_tests")
+
+@nox.session(name="user-tests")
```
agriyakhetarpal, line 46:
```suggestion
    session.run("pytest", "tests/generated_project_tests")


@nox.session(name="user-tests")
```

{{cookiecutter.project_name}}/noxfile.py
```diff
@@ -1,20 +1,66 @@
...
+                "sphinx-build",
+                "-b",
+                "html",
+                "-W",
+                "--keep-going",
+                ".",
+                "build/html/",
+            )
+
+@nox.session(name="generated-project-tests")
```
agriyakhetarpal, line 39:
```suggestion
                "build/html/",
            )


@nox.session(name="generated-project-tests")
```

I have a few similar suggestions – I think we should configure Ruff to lint and ask for two lines between methods.

{{cookiecutter.project_name}}/noxfile.py
```diff
@@ -1,20 +1,66 @@
...
+
+@nox.session(name="user-tests")
+def run_user_tests(session):
+    """Run the tests for testing user written tests"""
+    session.install("setuptools", silent=False)
+    session.install("-e", ".[dev]", silent=False)
+    session.run("pytest", "tests/user_tests")
+
+@nox.session(name="coverage")
+def run_coverage(session):
```
agriyakhetarpal, line 54:
```suggestion
    session.run("pytest", "tests/user_tests")


@nox.session(name="coverage")
def run_coverage(session):
```

{{cookiecutter.project_name}}/noxfile.py
```diff
@@ -1,20 +1,66 @@
...
+    """Run the coverage tests and generate an XML report."""
+    session.posargs.append("--cov=src/pybamm_cookiecutter")
+    session.posargs.append("--cov-report=xml")
+    run_user_tests(session)
+
+@nox.session(name="dev")
```
agriyakhetarpal, line 60:
```suggestion
    session.posargs.append("--cov-report=xml")
    run_user_tests(session)


@nox.session(name="dev")
```

{{cookiecutter.project_name}}/tests/generated_project_tests/test_entry_points.py
```diff
@@ -0,0 +1,45 @@
...
+    """Testing if the values get loaded via parameter entry points and are equal when loaded through entry points"""
+    # Loading parameter_sets through entry points
+    parameters = {{ cookiecutter.__project_slug }}.parameter_sets['Chen2020']
+    # Loading parameter sets through the source file by dynamically loading Chen2020.py as a module
+    spec = importlib.util.spec_from_file_location("Chen2020mod", "src/{{ cookiecutter.__project_slug }}/parameters/input/Chen2020.py")
+    chen_module = importlib.util.module_from_spec(spec)
+    sys.modules["Chen2020mod"] = chen_module
+    spec.loader.exec_module(chen_module)
+    parameters_from_file = chen_module.get_parameter_values()
+    assert parameters.keys() == parameters_from_file.keys(), f"The keys in the module and local input file are not the same, expected {parameters.keys} got {parameters_from_file.keys()}"
```
agriyakhetarpal, line 29:
```suggestion
    assert parameters.keys() == parameters_from_file.keys(), f"The keys in the module and local input file are not the same, expected {parameters.keys}, got {parameters_from_file.keys()}"
```

{{cookiecutter.project_name}}/tests/generated_project_tests/test_entry_points.py
```diff
@@ -0,0 +1,45 @@
...
+    """Test if the models via entry points are loaded correctly."""
+
+    entry_points = list({{ cookiecutter.__project_slug }}.models)
+    models = Path("src/{{ cookiecutter.__project_slug }}/models/input/").glob("*.py")
+    # Making a list Parameter sets in the parameters/input directory
+    models = [x.stem for x in models]
+
+    assert models == entry_points, "Entry points missing either in pyproject.toml or in the input directory"
+
+def test_model_entry_point_load():
```
agriyakhetarpal, line 41:
```suggestion
    assert models == entry_points, "Entry points missing either in pyproject.toml or in the input directory"


def test_model_entry_point_load():
```

.github/workflows/test_on_push.yml
santacodes: I was using `pipx` as a temporary fix for failing `generated-project-tests`, the CI was returning `nox command not found` when the working directory was changed, it is probably a bug because I found no trace of a similar issue on nox's github page or any threads related to github-actions. [Here](https://github.com/pybamm-team/pybamm-cookiecutter/actions/runs/10125311771/job/28000668991?pr=29) is the failed CI logs of the most recent commit where I tried executing the nox installation commands after changing the working directory. [Here](https://github.com/pybamm-team/pybamm-cookiecutter/actions/runs/10124950232/job/27999920383) are the logs for the previous CI steps where nox installation and changing the working directory where two different steps in the CI. I am not sure how to proceed with this issue. I tried the `uvx` experimental build to run nox session locally, but that doesn't seem to work in CI as well.

santacodes: Okay, so I got it working by using [nox's github-action](https://github.com/wntrblm/nox/blob/main/action.yml) `wntrblm/nox@2024.04.15` to setup and install nox. It uses `uv` by default to install the packages so it also preserves uniformity in the workflow.

{{cookiecutter.project_name}}/tests/generated_project_tests/test_entry_points.py
```diff
@@ -0,0 +1,46 @@
+import pytest
```
brosaplanella, line 1: Is this a duplicate of `tests/project_tests/test_entry_points.py`? If so, is there a way to have the file only once?

brosaplanella (commented):
I just have a small comment about the test duplication.

{{cookiecutter.project_name}}/tests/generated_project_tests/test_entry_points.py
santacodes: Yes, this is a duplicate file of the original `test_entry_points.py` in the project, it ensures that the entry points are working well within a `generated project` using a template. We had a discussion previously to move most of the things out of the project into the template except the `project generation tests` and `publishing CI`. In the long run, we wouldn't be maintaining or testing the project much, rather, we would be testing the template, i.e. generating a project using various kinds of build backends and such and testing the units inside them, because at the end of the day, we are going to deliver the template to the users, so testing the project doesn't make much sense. So we would be removing the copy of `test_entry_points.py` from the project and keeping it in the template, if that makes sense.

Also in the previous meeting, we discussed about providing a CLI with the template, so that a user doesn't have to go through a hassle to figure things out, just one command in the command line would set up a project in the current working directory. This would also make it easy to publish things to PyPI, manage releases, and support versions of PyBaMM(because some of the code within the template is derived from PyBaMM so we might need to update things as things get updated upstream). So to sum it up, 
```bash
pip install pybamm-cookiecutter
pybamm-cookiecutter
```
would basically initialise an empty project in the current working directory of your terminal.

brosaplanella (approved):

{{cookiecutter.project_name}}/noxfile.py
```diff
@@ -1,20 +1,70 @@
...
+                "build/html/",
+            )
+
+
+@nox.session(name="generated-project-tests")
+def run_generated_project_tests(session):
+    """Run the tests for testing units inside generated project"""
+    session.install("setuptools", silent=False)
+    session.install("-e", ".[dev]", silent=False)
+    session.run("pytest", "tests/generated_project_tests")
```
arjxn-py, line 45: Feel free to define a function as this is repeating - 
```python
def install_and_run_tests(session, test_dir):
    """Install dependencies and run tests in the specified directory."""
    session.run("pytest", test_dir)
```
And use it as - `install_and_run_tests(session, "tests/template_tests")` inside sessions.

{{cookiecutter.project_name}}/noxfile.py
```diff
@@ -1,20 +1,70 @@
...
+
+
+@nox.session(name="user-tests")
+def run_user_tests(session):
+    """Run the tests for testing user written tests"""
```
arjxn-py, line 50:
```suggestion
    """Run user written tests"""
```
Nit-picking :)

noxfile.py
```diff
@@ -36,12 +36,20 @@ def build_docs(session: nox.Session) -> None:
                 "build/html/",
             )
 
-@nox.session(name="test-generation")
+@nox.session(name="template-tests")
 def run_template_generation(session):
     """Run the tests for testing template generation"""
```
arjxn-py, line 41:
```suggestion
    """Run tests for the template generation."""
```

noxfile.py
```diff
@@ -36,12 +36,20 @@ def build_docs(session: nox.Session) -> None:
...
     session.install("setuptools", silent=False)
     session.install("-e", ".[dev]", silent=False)
-    session.run("pytest", "tests")
+    session.run("pytest", "tests/template_tests")
+
+
+@nox.session(name="project-tests")
+def run_project_tests(session):
+    """Run the tests for testing template generation"""
```
arjxn-py, line 49: Identical docstring as `template-tests`?

hooks/post_gen_project.py
```diff
@@ -20,5 +20,4 @@ def prepare_git() -> None:
     )
 
 if __name__ == "__main__":
-    print("Executing script!!")
     prepare_git()
```
arjxn-py, line 23: Instead of printing we may add some logging above i.e. - 
```python
import subprocess
import logging

logging.basicConfig(level=logging.INFO)

def prepare_git() -> None:
    git_author_name = "{{cookiecutter.full_name}}"
    git_author_email = "{{cookiecutter.email}}"
    git_branch = "{{cookiecutter.branch}}"

    commands = [
        ["git", "init", "-b", git_branch],
        ["git", "config", "user.name", git_author_name],
        ["git", "config", "user.email", git_author_email]
    ]

    for command in commands:
        try:
            subprocess.run(command, check=True)
            logging.info(f"Ran command: {' '.join(command)}")
        except subprocess.CalledProcessError as e:
            logging.error(f"An error occurred while running command: {' '.join(command)}. Error: {e}")
            break

if __name__ == "__main__":
    logging.info("Preparing git...")
    prepare_git()
    logging.info("Git setup complete.")
```

noxfile.py
santacodes: Oops yeah will fix that!

hooks/post_gen_project.py
santacodes: I think this file will be irrelevant once we migrate to `copier` in PR #34 as I removed the hooks directory and this file and added these procedures under `_tasks:` key in `copier.yml`. Overall copier has better support for inbuilt exception handling so we should be good with that I guess?

.github/workflows/test_on_push.yml
```diff
@@ -80,10 +86,52 @@ jobs:
...
+          with:
+            uv-venv: ".venv"
+
+        - name: Install nox
+          uses: wntrblm/nox@2024.04.15
+
+        - name: Test the generated project
+          working-directory: ./pybamm-example-project
+          run: |
+            nox -s generated-project-tests
```
Saransh-cpp, line 137:
```suggestion
          run: nox -s generated-project-tests
```

cookiecutter.json
```diff
@@ -40,7 +42,8 @@
         "license": "Select a license",
         "backend": {
             "__prompt__": "Choose a build backend",
-            "hatch": "Hatchling (recommended for pure Python projects)"
+            "hatch": "Hatchling (recommended for pure Python projects)",
+              "vcs": "Let hatch get the version from your version control system (Git/Mercurial/SVN)?"
```
Saransh-cpp, line 46: More scalable language
```suggestion
              "vcs": "Obtain the version from your version control system (Git/Mercurial/SVN)?"
```

{{cookiecutter.project_name}}/pyproject.toml
```diff
@@ -2,7 +2,9 @@
 {%- if cookiecutter.backend == "hatch" %}
 requires = [
   "hatchling",
-  "hatch-vcs"
+{%- if cookiecutter.backend == "hatch" and cookiecutter.vcs == true %}
```
Saransh-cpp, line 5: Already inside the `if cookiecutter.backend == "hatch"` block
```suggestion
{%- if cookiecutter.vcs == true %}
```

{{cookiecutter.project_name}}/pyproject.toml
```diff
@@ -86,12 +88,19 @@ SPM = "{{ cookiecutter.__project_slug }}.models.input.SPM:SPM"
 {# keep this line here for newline #}
 {%- if cookiecutter.backend == "hatch" %}
 [tool.hatch]
+{%- if cookiecutter.backend == "hatch" and cookiecutter.vcs == true %}
```
Saransh-cpp, line 91:
```suggestion
{%- if cookiecutter.vcs == true %}
```

{{cookiecutter.project_name}}/src/{{cookiecutter.__project_slug}}/__init__.py
```diff
@@ -16,7 +20,9 @@
 {%- else %}
 __all__ = [
 {%- endif %}
+{%- if cookiecutter.vcs %}
     "__version__",
+{%- endif %}
```
Saransh-cpp, line 25: Should always be there
```suggestion
    "__version__",
```

Saransh-cpp (changes requested):

Saransh-cpp (approved):
Thanks, @santacodes! This looks great now!

PR #30, opened 2024-07-22
Added Basic Reservoir Model

santacodes:
+ Added `basic_reservoir.py` adopted from Ferran's draft PR - pybamm-team/PyBaMM/pull/4184.
+ Added the `model entry` in `pyproject.toml`.
+ Changed the file naming conventions of models to abide by the ones in `PyBaMM`.

Subtask - #26


Saransh-cpp (approved):
Thanks, @santacodes! Looks good to me, but I'll wait for a review from @brosaplanella.

pyproject.toml
```diff
@@ -65,7 +65,8 @@ Changelog = "https://github.com/pybamm-team/pybamm-cookiecutter/releases"
 Chen2020 = "pybamm_cookiecutter.parameters.input.Chen2020:get_parameter_values"
 
 [project.entry-points."models"]
-SPM = "pybamm_cookiecutter.models.input.SPM:SPM"
+SPM = "pybamm_cookiecutter.models.input.spm:SPM"
+Reservoir = "pybamm_cookiecutter.models.input.basic_reservoir:BasicReservoir"
```
Saransh-cpp, line 69: Do we even need these in the project now?

santacodes: I think we could start removing things from the project when things inside the template get stable. At the end of the day, we will deliver the template to the users and not the project, so there is not much of a reason to maintain and test the project, other than for the template generation tests and maybe prototyping.

Saransh-cpp: Sounds good, thanks!

brosaplanella (approved):

pyproject.toml
agriyakhetarpal, line 69:
```suggestion
SPM = "pybamm_cookiecutter.models.input.spm:SPM"
BasicReservoir = "pybamm_cookiecutter.models.input.basic_reservoir:BasicReservoir"
```

{{cookiecutter.project_name}}/pyproject.toml
```diff
@@ -81,7 +81,8 @@ Changelog = "{{ cookiecutter.url }}/releases"
 Chen2020 = "{{ cookiecutter.__project_slug }}.parameters.input.Chen2020:get_parameter_values"
 
 [project.entry-points."models"]
-SPM = "{{ cookiecutter.__project_slug }}.models.input.SPM:SPM"
+SPM = "{{ cookiecutter.__project_slug }}.models.input.spm:SPM"
+Reservoir = "{{ cookiecutter.__project_slug }}.models.input.basic_reservoir:BasicReservoir"
```
agriyakhetarpal, line 85:
```suggestion
BasicReservoir = "{{ cookiecutter.__project_slug }}.models.input.basic_reservoir:BasicReservoir"
```

agriyakhetarpal (approved):
Looks great! I hope my suggestions below are in line with what you discussed you had to do in today's meeting.

{{cookiecutter.project_name}}/pyproject.toml
arjxn-py: Just merged your suggestions hoping that it'd fix the failures

santacodes: To fix the failures we would have to change the model names to the same ones as their keys in entry points. I will do it real quick!

santacodes: It should work now, the assertion was failing because there was a list comparison and the lists weren't sorted so the unordered elements were causing the assertion to fail.

PR #32, opened 2024-07-23
Added Git Initialisation Hooks

santacodes:
+ Added `git` initialisation hooks in the template to automatically initialise a `git repository` when a `generated project` is created by `cookiecutter`.
+ Added a default `main` branch parameter for git repository initialisation.

Subtask - #26

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -6,6 +6,7 @@
     ],
     "org": "pybamm-team",
     "url": "https://{{ cookiecutter.platform }}.com/{{ cookiecutter.org }}/{{ cookiecutter.project_name }}",
+    "branch": "main",
     "full_name": "John Doe",
     "email": "john@doe.com",
     "project_short_description": "A template for creating battery modeling projects based on PyBaMM",
@@ -31,7 +32,8 @@
             "gitlab": "GitLab"
         },
         "org": "The name of your organisation (or username, if you are not part of an organisation)",
-        "url": "The URL to your repository",
+        "url": "The Origin URL to your repository",
+        "branch": "The default branch of your repository",
         "full_name": "Your name",
         "email": "Your email",
         "project_short_description": "A short description of your project",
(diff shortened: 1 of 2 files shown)


Saransh-cpp (approved):
Looks good, thanks @santacodes! I haven't tested this locally but could you confirm if this is working locally?

agriyakhetarpal (approved):
Thanks, @santacodes! While I shall approve this PR now so that it gets your other PR going, I would suggest that we provide an option in the template to not require `hatch-vcs` by default. Making this it an opt-in piece of functionality would be great, this way, this script would not be needed for installing the package. Moreover, I feel that new users of PyBaMM and newcomers to Python packaging would like to hardcode the package version in a string instead of trying to wrap their heads around version control systems.

santacodes:
> Looks good, thanks @santacodes! I haven't tested this locally but could you confirm if this is working locally?

Yes it does, I tested it locally before pushing it.

arjxn-py:
I'm also testing it quick locally

santacodes:
> Thanks, @santacodes! While I shall approve this PR now so that it gets your other PR going, I would suggest that we provide an option in the template to not require `hatch-vcs` by default. Making this it an opt-in piece of functionality would be great, this way, this script would not be needed for installing the package. Moreover, I feel that new users of PyBaMM and newcomers to Python packaging would like to hardcode the package version in a string instead of trying to wrap their heads around version control systems.

I completely agree, I will add a new prompt for that to opt-in `hatch-vcs` and make it `false` by default so that they don't have to go through that hassle. Perhaps as you said the versions could be hardcoded in the `_version.py` file and can be configured through `tool.hatch.version`. I am working on the patch and will push that commit in #29, it should possibly resolve all the failing tests as well. Thank you for the insights!

arjxn-py (approved):
Works well, I'd now need to re-configure my git name & mail 😆

PR #36, opened 2024-08-02
Added `pybamm-cookiecutter` CLI 

santacodes:
## Additions
+ Added a basic `cli` with an optional `--path` argument defining the path generated project should reside within.
+ Added basic documentation for `cli` usage in `cli.py` and `README.md`.

## Removals
- Removed `models`, `entry points`, and `parameter sets` from the project as they are now populated in the template.
- Removed the `project-tests` from `noxfile` and updated workflow to only test the template.
- Removed the entry points from `pyproject.toml` and `pybamm` as a dependency in the project.

## Usage 
To test this, check out this branch and inside the repository and do a local `pipx` installation on your machine.
```bash 
git clone https://github.com/santacodes/pybamm-cookiecutter -b cli
cd pybamm-cookiecutter/
pipx install . 
```
After installation, the `CLI` will be available systemwide and can be accessed using the `pybamm-cookiecutter` command.
For help references add the `-h` or `--help` flag, e.g. `pybamm-cookiecutter --help` which would prompt with all the available arguments.

Sub-task #26


README.md
```diff
@@ -43,7 +45,30 @@ Generate a project from the `pybamm-cookiecutter` template.
...
+pipx install .
+```
+Once `pipx` installs the package locally, you can execute it to generate a project.
+```bash
+pybamm_cookiectter
+# this will prompt the user for project configuration and generate
+# the project in the current working directory.
+# or
+pybamm_cookiecutter --path /path_to_copy_to
+# this will install the project in the specified path.
```
agriyakhetarpal, line 71: I think we should recommend generating the project in a separate directory from the first go since running it in the current working directory (where the `pybamm-cookiecutter` project exists) will raise many conflicts. Either way, I think we should remove the git clone step and assume that we are writing the documentation in a way that suggests that `pybamm-cookiecutter` is already on PyPI, given that we are publishing it pretty soon.

README.md
```diff
@@ -43,7 +45,30 @@ Generate a project from the `pybamm-cookiecutter` template.
 **Note:** This requires an internet connection. You could manually clone the git repository and run the copy command, or just execute the copy command with the URL to the git repository.
```
agriyakhetarpal, line 46: I would suggest removing this as well, in addition to the comment below

pyproject.toml
```diff
@@ -55,19 +55,18 @@ docs = [
   "sphinx-gallery",
 ]
 
+[project.scripts]
+pybamm_cookiecutter = "pybamm_cookiecutter.cli:pybamm_cookiecutter_cli"
+
+[project.entry-points."pipx.run"]
+pybamm_cookiecutter = "pybamm_cookiecutter.cli:pybamm_cookiecutter_cli"
+
```
agriyakhetarpal, line 63:
```suggestion
[project.scripts]
pybamm-cookiecutter = "pybamm_cookiecutter.cli:pybamm_cookiecutter_cli"

[project.entry-points."pipx.run"]
pybamm-cookiecutter = "pybamm_cookiecutter.cli:pybamm_cookiecutter_cli"

```

I think these should both be `pybamm-cookiecutter`, similar to how other packages that offer CLIs with hyphens in their names have been packaged and made available through `pipx` – hyphens are accepted in this case, and having an underscore doesn't seem necessary.

src/pybamm_cookiecutter/cli.py
```diff
@@ -0,0 +1,40 @@
+import copier
+import argparse
+from pathlib import Path
+from colorama import Fore
+import os
```
agriyakhetarpal, line 5: Would we need to add `colorama` as a dependency?

agriyakhetarpal (commented):
This looks nice, thanks, @santacodes! Since the files (for the models and the entry point handler) from the project have been removed, I think we can remove references to them from our bundled licenses (and add them to the generated project's bundled licenses if they are not present).

README.md
santacodes: No, that will not be the case with the CLI when installed through `pipx`, it is installed system-wide and hence there won't necessarily be a case where a user generates the project inside the `pybamm-cookiecutter` project unless they try to install it from the source and execute the command while being inside the project directory unknowingly. 
In most cases, a user would perform - 
```bash 
pipx install pybamm-cookiecutter
# Assuming that I am in my home directory `/home` and I give the project name pybamm-test, it will look something like this - 
pybamm-cookiecutter # user prompt initialises and I give the project name as mentioned above
```
`pybamm-test` will be generated like `home/pybamm-test/src/pybamm_test/...`, but obviously it is up to the user to provide the `--path` argument as well.
I think your main concern is executing this command in an unknown directory and all of the project files populating themselves in that directory creating a mess which is why I added a directory restructure within this PR which would automatically create a directory with the `project_name` and populate all the files inside it, so if it is a mistake a user can easily just delete the parent directory, in this case `pybamm-test`. 

In case they opt to provide the path argument and want to have a generated project inside the `home/test/` path, the generated project path would look something like this -
```bash 
pybamm-cookiecutter --path /home/test 
```
The generated project path would be `home/test/pybamm-test/src/pybamm_test`.

I think the documentation part I agree with writing it in the way which resembles the behaviour of a `pipx` `pybamm-cookiecutter` package.  I would love to hear what @brosaplanella thinks as well as I mostly added the documentation for immediate setup purposes for his projects.

src/pybamm_cookiecutter/cli.py
agriyakhetarpal: If we are using it in a file, we should still list it in our dependencies since we can't guarantee whether `copier` will continue to depend on it.

README.md
agriyakhetarpal: Ah, apologies if I was too terse, let me clarify:

> No, that will not be the case with the CLI when installed through `pipx`, it is installed system-wide and hence there won't necessarily be a case where a user generates the project inside the `pybamm-cookiecutter` project

Indeed, the usage instructions were instructing the user to run it with `pipx` in the git-clone. You have now updated this in da1e56666f4e452c73e58284af28b2e36b2520c7, which is fine with me.

> I think your main concern is executing this command in an unknown directory and all of the project files populating themselves in that directory creating a mess

Yes, I was under the impression that a user would execute the command either inside the directory where they want their project, or that they would let the command create the directory. But, I don't expect a user to run it inside a non-empty directory in the first place, so there shouldn't be any conflicts with your new approach.

> which is why I added a directory restructure within this PR which would automatically create a directory with the project_name and populate all the files inside it

Sounds good!

My only point of contention was that developers should not use `pipx` to install the project from source (whether in editable mode or not); a standard `pip install` would be better – this is already covered by `nox`, so there's nothing more to add.

The other point was that we are quite close to the PyPI release, and therefore, we should remove notes about the project not being on PyPI yet in the paragraph above this one.

santacodes: I missed the directory restructure commit in my last `copier` migration PR which would have avoided all the confusion haha and should've made it clearer, my mistake apologies! 😅

agriyakhetarpal: No worries, thanks for fixing that 🙏🏻

README.md
```diff
@@ -31,19 +31,31 @@ This repository contains a `copier` template for battery modeling projects using
...
-**Note:** This requires an internet connection. You could manually clone the git repository and run the copy command, or just execute the copy command with the URL to the git repository.
 
 ```bash
-copier copy https://github.com/pybamm-team/pybamm-cookiecutter.git name_of_your_project/ --trust
+copier copy https://github.com/pybamm-team/pybamm-cookiecutter.git . --trust
+# this will generate the project in the current working directory
+copier copy https://github.com/pybamm-team/pybamm-cookiecutter.git path_to_copy_to/ --trust
+# this will generate the project in the specified path
+```
+#### Using pipx
```
agriyakhetarpal, line 50:
```suggestion
#### Using pipx (recommended)
```

Maybe we can add a small sentence below this heading as well and link to https://pipx.pypa.io/stable/

README.md
```diff
@@ -31,19 +31,31 @@ This repository contains a `copier` template for battery modeling projects using
...
+
+Install the `pybamm-cookiecutter` package.
+```bash
+pipx install pybamm-cookiecutter
+```
+Navigate into the directory you want your project directory to reside in, or use `--path` argument to explicitly mention the path where you want your project to be generated.
+```bash
+pybamm-cookiecutter --path /path_to_copy_to
 ```
```
agriyakhetarpal, line 59: We should mention that installing the package with `pipx` is not necessary, even though it is possible to install the package. The command `pipx run pybamm-cookiecutter <args...>` is equally fine.

src/pybamm_cookiecutter/__init__.py
```diff
@@ -5,15 +5,10 @@
 """
 from __future__ import annotations
 
-import pybamm
-
 from ._version import version as __version__
-from .entry_point import Model, parameter_sets, models
+from .cli import pybamm_cookiecutter_cli
```
agriyakhetarpal, line 9: Could you change the relative imports here to absolute imports?

src/pybamm_cookiecutter/__main__.py
```diff
@@ -0,0 +1,3 @@
+if __name__ == "__main__":
+    from pybamm_cookiecutter.cli import pybamm_cookiecutter_cli
+    pybamm_cookiecutter_cli()
```
agriyakhetarpal, line 3:
```suggestion
    from pybamm_cookiecutter.cli import pybamm_cookiecutter_cli

    pybamm_cookiecutter_cli()
```

src/pybamm_cookiecutter/cli.py
```diff
@@ -0,0 +1,40 @@
...
+    -------
+    $ pybamm-cookiecutter
+    Generates a project in the current working directory of the terminal.
+    $ pybamm-cookiecutter --path /myproject
+    Generates a project in the `myproject` directory.
+    """
+    try:
+        parser = argparse.ArgumentParser(description = "A copier template generator for PyBaMM based projects")
+        parser.add_argument("--path", type = str, required = False, default = os.getcwd(),
+                            help = "The destination path for project generation")
```
agriyakhetarpal, line 27:
```suggestion
                            help = "The destination path for project generation. The default is the current working directory")
```

Saransh-cpp:
@santacodes could you merge main here?

santacodes:
> @santacodes could you merge main here?

I did merge it, I think the merge commit got hidden within the `resolved merge conflicts` commit [here](https://github.com/pybamm-team/pybamm-cookiecutter/commit/d49e1bae6ab671481a009f5602afe34cf97ac388). Sorry if it didn't show up, I should have probably rebased it but I resolved the merge conflicts locally and pushed them here as they were a bit complicated to resolve and there was no web editor option with this one.

agriyakhetarpal (commented):
This is looking good, thanks, @santacodes! Here are some of my suggestions besides the code review that are better provided in text:

1. Installing and running it displays the message `No git tags found in template, using HEAD as ref` – this is marked in red, which might prompt users/newcomers to think that something is wrong. Is there a way to disable this message or make it clearer? That said, I don't think this is coming from any of our changes, though.
2. The CLI has a problem where it truncates long sentences, at least on macOS – I noticed that the sentence was terminated before completion as follows: "such as `import p"
4. There's no option to go back if I made an error while typing things out – does `copier` support undo functionality?
5. I think we need better exception handling. If I terminate the program in between by pursuing a `KeyboardInterrupt` (Command + D on my machine), it prints out "Error caused by an exception:" followed by a long list of metadata that was being used to generate the project, stopped midway. Comparing this with the `copier` CLI, this is just "Execution stopped by user", which is more elegant. Also, Is there a way to distinguish a keyboard interrupt from an actual error?
6. Also, in the previous point, a keyboard interrupt midway into creating a project returns an exit code of 0, indicating success, instead of a non-zero exit code to indicate failure – I think it should be 0 for successful completion, 1 for an error in the project generation, and 2 for a keyboard interrupt, and higher for any other non-standard error.

.github/workflows/test_on_push.yml
```diff
@@ -136,6 +159,14 @@ jobs:
...
+      - name: Install copier and jinja2_time and generate a template with default values
+        run: |
+          uv pip install copier jinja2-time
+          copier copy . . --trust --defaults
+
+      - name: Check if the documentation can be built
+        working-directory: ./pybamm-example-project
+        run: |
+          git add .
+          git commit -am "initial commit"
```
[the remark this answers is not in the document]
santacodes: Yes, that was discussed in the last PR when Ferran raised it up, you could refer it [here](https://github.com/pybamm-team/pybamm-cookiecutter/pull/39#discussion_r1704192182).

santacodes:
> This is looking good, thanks, @santacodes! Here are some of my suggestions besides the code review that are better provided in text:
> 
> 1. Installing and running it displays the message `No git tags found in template, using HEAD as ref` – this is marked in red, which might prompt users/newcomers to think that something is wrong. Is there a way to disable this message or make it clearer? That said, I don't think this is coming from any of our changes, though.

I think this could be resolved by providing a `vcs_ref` via the command line or the copier API, I will look into that.
> 2. The CLI has a problem where it truncates long sentences, at least on macOS – I noticed that the sentence was terminated before completion as follows: "such as `import p"

Could you please attach a screenshot of some kind it's hard for me to picture it 😅 .
> 3. There's no option to go back if I made an error while typing things out – does `copier` support undo functionality?

I think it does not, looking at the documentation. Maybe I could try to come up with some kind of a hacky way to do it something like recording a particular keystroke for example `ctrl+z` and then with the cached recorded user inputs, generate the project using the copier API again which would probably continue from where the user left it off. I think that would be something unreliable though, but if we are doing it, this might just be one of the ways to do it.

> 4. I think we need better exception handling. If I terminate the program in between by pursuing a `KeyboardInterrupt` (Command + D on my machine), it prints out "Error caused by an exception:" followed by a long list of metadata that was being used to generate the project, stopped midway. Comparing this with the `copier` CLI, this is just "Execution stopped by user", which is more elegant. Also, Is there a way to distinguish a keyboard interrupt from an actual error?

I agree with that, I will add better exception handling in the CLI and try to cover most of the endpoints.

> 5. Also, in the previous point, a keyboard interrupt midway into creating a project returns an exit code of 0, indicating success, instead of a non-zero exit code to indicate failure – I think it should be 0 for successful completion, 1 for an error in the project generation, and 2 for a keyboard interrupt, and higher for any other non-standard error.

For that, I think we could just wrap the exception handling to `KeyBoardInterrupt` rather than exit codes.

agriyakhetarpal:
Yes, letting Python handle the `KeyBoardInterrupt` will let it use its related exit code, I think. Let's skip the undo functionality for now (or you could try your idea – if the implementation is minimal, that's great).

For the screenshot, here you go:

pybamm-cookiecutter command-line usage example in order to reproduce an error with the truncation of a specific sentence

which indicates that the prompt for the second question was cut off. Maybe there's a limit on the number of characters in that can be configured, or maybe it's a hard limit and we can reword the sentence to make it shorter?

agriyakhetarpal:
Also, if we were to run `copier` internally in a `subprocess.call()` – i.e., `sys.exit(subprocess.call(argv))`, that should resolve the problem with the exit codes automatically. We'll have to switch to using the `copier` CLI instead of the API, though.

santacodes:
> Also, if we were to run `copier` internally in a `subprocess.call()` – i.e., `sys.exit(subprocess.call(argv))`, that should resolve the problem with the exit codes automatically. We'll have to switch to using the `copier` CLI instead of the API, though.

I think it might be a bad idea to use the `subprocess` library to access the copier CLI, the exit code 0 on `KeyboardInterrupt` is predefined [here](https://github.com/copier-org/copier/blob/03156740b511ed752c7af53d9895da3e09480770/copier/cli.py#L80) in the copier code base even for the CLI. Maybe we should aim for a better exception handling in the `pybamm-cookiecutter` CLI to cover all the cases.

agriyakhetarpal:
I would say we should use what brings simpler code to write (and test) – my suggestion is not about handling `KeyboardInterrupt`s, since there are other exceptions to handle. If a subprocess call for us can directly inherit this behaviour from `copier`, we might as well do so (since `copier` is not going to deprecate its CLI anytime soon).

santacodes:
> I would say we should use what brings simpler code to write (and test) – my suggestion is not about handling `KeyboardInterrupt`s, since there are other exceptions to handle. If a subprocess call for us can directly inherit this behaviour from `copier`, we might as well do so (since `copier` is not going to deprecate its CLI anytime soon).

Umm I think you misunderstood me, you should perhaps look at this [copier CLI](https://github.com/copier-org/copier/blob/master/copier/cli.py) source code, but in brief, the CLI handles only three exceptions out of which 2 are copier specific errors which are already handled in the `Worker` class. So basically how all this works is the `copier CLI` takes in all the arguments from the command line, parses it and passes the command line arguments to a `Worker` instance, and does the `run_copy()` method [here](https://github.com/copier-org/copier/blob/03156740b511ed752c7af53d9895da3e09480770/copier/cli.py#L279). In our CLI we are performing the same thing, except we are directly performing the `run_copy` method of the `Worker` class like defined [here](https://github.com/copier-org/copier/blob/master/copier/main.py#L1127) in the source code. So the user prompts which you are concerned about are actually generated by the `Worker` class and not the `copier CLI`, which is also why we see the same prompt UI when we use the `copier` command or access it through the `run_copy()` command. I just feel we would just increase a layer of abstraction with a subprocess call just for a `KeyboardInterrupt` exception because right now the `pybamm-cookiecutter CLI` works like -

`pybamm-cookiecutter CLI` -> `Worker.run_copy`
while your proposed method would do something like -
`pybamm-cookiecutter CLI` -> `copier CLI` -> `Worker.run_copy`

And even the word trimmings you mentioned earlier shouldn't be much of an issue because they only get trimmed after a user answers the prompt if I am not wrong, I tested that locally as well. Overall, I think we should be good with handling the `KeyboardInterrupt` error in our CLI as that might be the only error we might run into from our side because the `Worker` class directly handles every other exception case in the core library and not the `copier CLI`. I hope my explanation made sense😅, feel free to correct me haha.

agriyakhetarpal:
Thanks for the explanation and the cross-links to the code, @santacodes – makes sense to me!

agriyakhetarpal:
Thanks for the explanation and the cross-links to the code, @santacodes – makes sense to me! Yes, there's no need for extra abstractions, then. Please let us know when this is ready for review again.

.github/workflows/test_on_push.yml
```diff
@@ -47,6 +47,11 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Set up Git
+        run: |
+          git config --global user.name "pybamm user"
+          git config --global user.email "pybammuser@pybamm.org"
+
```
agriyakhetarpal, line 54: Nitpicky comment: could you move this step so that it is just before where the tests are run? That makes it a little clearer as to what the step is required for – and that the `uv` and `nox` installation steps are not related to this.

.github/workflows/test_on_push.yml
```diff
@@ -121,12 +122,39 @@ jobs:
...
           python-version: ${{ matrix.python-version }}
 
+      - name: Set up uv
+        uses: yezz123/setup-uv@v4
+        with:
+          uv-venv: ".venv"
+
       - name: Install copier and jinja2_time
-        run: python -m pip install copier jinja2_time
+        run: uv pip install copier jinja2_time
```
agriyakhetarpal, line 131:
```suggestion
        run: uv pip install copier jinja2-time
```

noxfile.py
```diff
@@ -39,18 +39,14 @@ def install_and_run_tests(session, test_dir):
     """Install dependencies and run tests in the specified directory."""
     session.install("setuptools", silent=False)
     session.install("-e", ".[dev]", silent=False)
+    session.run("pipx", "install", ".", "--force", silent=False)
```
agriyakhetarpal, line 42: I don't think it is a good idea to use `pipx` inside a `nox` session, is there an alternate – could we replace this with a standard install? Also, GHA runners come with `pipx` installed, but this might not be the same for local development, so if we were to do this, we would need to install it with `session.install("pipx")`.

pyproject.toml
```diff
@@ -55,26 +56,29 @@ docs = [
...
 version.source = "vcs"
 build.hooks.vcs.version-file = "src/pybamm_cookiecutter/_version.py"
 envs.default.dependencies = [
   "pybamm",
 ]
 
+[tool.hatch.build.targets.wheel]
+only-include = ["src/pybamm_cookiecutter", "/template", "copier.yml"]
+sources = ["src"]
+
```
agriyakhetarpal, line 81: Does this configuration include the LICENSE and the bundled licenses? Could you unpack both the sdist and wheel and verify? If it doesn't, we should include those files as well.

src/pybamm_cookiecutter/_version.py
```diff
@@ -0,0 +1,16 @@
...
+else:
+    VERSION_TUPLE = object
+
+version: str
+__version__: str
+__version_tuple__: VERSION_TUPLE
+version_tuple: VERSION_TUPLE
+
+__version__ = version = '0.1.dev142+g0fc714a.d20240813'
+__version_tuple__ = version_tuple = (0, 1, 'dev142', 'g0fc714a.d20240813')
```
agriyakhetarpal, line 16: If the file isn't to be tracked in version control (as mentioned by the setuptools_scm auto-generated comment here), we should find another method to handle this.

pyproject.toml
santacodes: They don't, will add them!

src/pybamm_cookiecutter/_version.py
santacodes: I am not sure how we would go about handling this, maybe we should remove `hatch-vcs` from the project as we are anyways going to be following semantic versioning.

src/pybamm_cookiecutter/_version.py
```diff
@@ -0,0 +1,16 @@
+# file generated by setuptools_scm
```
Saransh-cpp, line 1: This file should not be tracked?

tests/template_tests/test_project_generation.py
```diff
@@ -33,6 +36,15 @@ def test_template_with_extra_answers(copie): # codespell:ignore copie
...
+    with open(result.project_dir / extra_context["project_name"] / "README.md") as f:
         assert f.readline() == f"# {extra_context['project_name']}\n", f"{f.readline()} is not the same as {extra_context['project_name']}\n"
+
+def test_cli():
+    """
+    Testing if the CLI works and returns a successful exit code on execution
+    """
+    os.mkdir("testcli")
+    return_code = subprocess.run(["pybamm-cookiecutter", "--defaults"], cwd = "./testcli")
+    shutil.rmtree("testcli")
```
Saransh-cpp, line 49: Files and folders of the generated project are being checked above so this minimal test should be fine ✅

.gitignore
```diff
@@ -127,8 +127,5 @@ cython_debug/
 # Code editor files and folders
 .vscode/
 
-# hatch-vcs
-src/pybamm_cookiecutter/_version.py
-
```
Saransh-cpp, line 132: Why was this removed?

Saransh-cpp (changes requested):
Thanks, @santacodes! This looks really good. A couple of comments below -

src/pybamm_cookiecutter/_version.py
Saransh-cpp: Just saw this. Semantic/calendar versioning choice is different from VCS/manual versioning choice. VCS versioning just makes our life easy. I am not sure why is this file required to be tracked by `git` in this PR, but we can discuss this in today's meeting!

santacodes: Makes sense, I am removing the file for now

Saransh-cpp (approved):
Thanks for working on this, looks amazing, @santacodes!

> Good practice tip: You should revert the commit where an unwanted file (in this case - `_version.py`) was added instead of deleting the file in another commit. Adding and deleting the file still leaves it in the git history. We don't require reverting here as we will "squash and merge" the PR, so the git history within this PR will be lost anyways.

Saransh-cpp:
Could you also add the CI badge in the README - [Test template and generated project](https://github.com/pybamm-team/pybamm-cookie/actions/workflows/test_on_push.yml)

src/pybamm_cookiecutter/cli.py
```diff
@@ -0,0 +1,60 @@
...
+        parser.add_argument(
+            '--version',
+            action='version',
+            version=f'PyBaMM Cookiecutter CLI Version - {version}'
+        )
+
+        parser.add_argument(
+            "--defaults",
+            action="store_true",
+            help="Whether to use default options for generating the template"
```
arjxn-py, line 44: Something like `pybamm-cookiecutter help` should be helpful, is it already there?

arjxn-py (commented):
Thanks @santacodes, a couple of comments below

src/pybamm_cookiecutter/cli.py
arjxn-py: if not can u add it?

agriyakhetarpal (commented):
Thanks, @santacodes! I tested out the CLI exceptions and this looks good to go now. I do have one suggestion, so I'm not approving this at the moment – happy to do so once it's fixed. It's about license compliance (xref #43) – I built the sdist and wheel, and while both do contain the relevant LICENSE files, this is the content of the project's (`pybamm-cookiecutter`'s) bundled LICENSES file, which doesn't seem to make much sense:

```
$ cat LICENSES-bundled.txt
```

```
This project and source distributions bundle several libraries that are
compatibly licensed.


Name: PyBaMM
Files: - src/{{ project_slug }}/parameters/input/Chen2020.py
       - src/{{ project_slug }}/models/input/BasicReservoir.py
       - src/{{ project_slug }}/models/input/SPM.py
       - src/{{ project_slug }}/entry_point.py
License: BSD-3-Clause
```

because `pybamm-cookiecutter` doesn't contain these files anymore, and `pybamm-cookiecutter` won't know about what `{{ project_slug }}` represents either until a project is generated. This should be moved to the generated project's `LICENSES-bundled.txt.jinja` file (which was outdated), since the generated project is where the files now exist. I think we can remove the `LICENSES-bundled.txt` file from `pybamm-cookiecutter` entirely, since we are now not using any third-party code.

src/pybamm_cookiecutter/cli.py
agriyakhetarpal: It's already there:

```bash
❯ pybamm-cookiecutter --help
```

```
usage: pybamm-cookiecutter [-h] [--path PATH] [--version] [--defaults]

A copier template generator for PyBaMM based projects

options:
  -h, --help   show this help message and exit
  --path PATH  The destination path for project generation. The default is the current working directory
  --version    show program's version number and exit
  --defaults   Whether to use default options for generating the template
```

src/pybamm_cookiecutter/cli.py
```diff
@@ -0,0 +1,60 @@
...
+        )
+        args = parser.parse_args()
+        destination_path = Path(args.path)
+
+        with copier.Worker(src_path = TEMPLATE, dst_path = destination_path, unsafe = True, defaults = args.defaults) as worker:
+            worker.run_copy()
+
+    except KeyboardInterrupt:
```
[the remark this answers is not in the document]
santacodes: Have a look at [this](https://copier.readthedocs.io/en/stable/reference/errors/#copier.errors.UserMessageError), it sounds like an exception class but it's just a method for fancy way of printing the exception rather.
So basically `UserMessageError` is just a fancy way of printing things with `colorama`.

agriyakhetarpal (approved):

agriyakhetarpal:
Thanks, @santacodes!

Issue #35, opened 2024-08-01
Issues when trying the template


brosaplanella:
I tried copying the template using copier (in WSL, Python 3.11) and I have spotted the following issues:
- User name needs to be passed between quotation marks, but this is not specified. Probably we could keep things as they are an pass the quotation marks ourselves when calling the `git config user.name` command.
- When calling nox (either `generated-project-tests` or `dev`) I get the following error:
```bash
nox > Running session dev
nox > Creating virtual environment (virtualenv) using python3.11 in .nox/dev
nox > python -m pip install virtualenv
nox > virtualenv /home/brosaplanella/LTES/encapsulated_LTES/venv
nox > /home/brosaplanella/LTES/encapsulated_LTES/venv/bin/python -m pip install -e '.[dev]'
nox > Warning: /home/brosaplanella/LTES/encapsulated_LTES/venv/bin/python is not installed into the virtualenv, it is located at /home/brosaplanella/LTES/encapsulated_LTES/venv/bin/python. This might cause issues! Pass external=True into run() to silence this message.
Obtaining file:///home/brosaplanella/LTES/encapsulated_LTES
ERROR: Exception:
Traceback (most recent call last):
  File "/home/brosaplanella/LTES/encapsulated_LTES/venv/lib/python3.11/site-packages/pip/_internal/cli/base_command.py", line 179, in exc_logging_wrapper
    status = run_func(*args)
             ^^^^^^^^^^^^^^^
  File "/home/brosaplanella/LTES/encapsulated_LTES/venv/lib/python3.11/site-packages/pip/_internal/cli/req_command.py", line 67, in wrapper
    return func(self, options, args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/brosaplanella/LTES/encapsulated_LTES/venv/lib/python3.11/site-packages/pip/_internal/commands/install.py", line 377, in run
    requirement_set = resolver.resolve(
                      ^^^^^^^^^^^^^^^^^
  File "/home/brosaplanella/LTES/encapsulated_LTES/venv/lib/python3.11/site-packages/pip/_internal/resolution/resolvelib/resolver.py", line 76, in resolve
    collected = self.factory.collect_root_requirements(root_reqs)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #39, opened 2024-08-05
Fixed documentation inside the template

santacodes:
As brought up in [this](https://github.com/pybamm-team/pybamm-cookiecutter/issues/35#issuecomment-2268953671) discussion the documentation should build now without any errors.


Saransh-cpp (approved):
This looks good to me, but I'll let @agriyakhetarpal (pybamm's goto documentation person, haha) do the final review

.github/workflows/test_on_push.yml
```diff
@@ -134,3 +134,39 @@ jobs:
...
+      - name: Install copier and jinja2_time and generate a template with default values
+        run: |
+          uv pip install copier jinja2_time
+          copier copy . ../pybamm-example-project --trust --defaults
+
+      - name: Check if the documentation can be built
+        working-directory: ../pybamm-example-project
+        run: |
+          git add .
+          git commit -am "initial commit"
```
brosaplanella, line 171: Is this needed?

template/docs/_static/pybamm.css
brosaplanella, line 1: Just to be clear, this was as a placeholder so I could push `_static` into Github, but the file itself is not needed. If there is a better way of doing it please go ahead.

template/noxfile.py.jinja
```diff
@@ -55,11 +56,12 @@ def run_user_tests(session):
 @nox.session(name="coverage")
 def run_coverage(session):
     """Run the coverage tests and generate an XML report."""
-    session.posargs.append("--cov=src/{{project_slug}}")
-    session.posargs.append("--cov-report=xml")
+    session.install("setuptools", silent=False)
+    session.install("coverage", silent=False)
+    session.install("-e", ".[dev]", silent=False)
+    session.run("pytest", "--cov=src/{{project_slug}}", "--cov-report=xml", "tests/user_tests")
     run_user_tests(session)
```
brosaplanella, line 63: I think this is not needed, as things stand. Things can probably be refactored to reuse code, my implementation was a quick and dirty fix to get things running.

brosaplanella (approved):
Looks good, just see a few comments

.github/workflows/test_on_push.yml
santacodes: Without initalising a commit the doctests seem to fail as HEAD is detached, probably an issue of `sphinx-last-updated-by-git` as it detects version history for builds. The exact logs are [here](https://github.com/pybamm-team/pybamm-cookiecutter/actions/runs/10249890182/job/28354413186).

template/docs/_static/pybamm.css
santacodes: I added it for the same reason, perhaps we could craft a logo of `pybamm-cookiecutter`? or just add the `css` and make it opt-in by not including it in `conf.py`?

template/docs/_static/pybamm.css
```diff
@@ -0,0 +1,230 @@
...
+
+.tooltipster-sidetip.tooltipster-shadow.tooltipster-left
+  .tooltipster-arrow-border {
+  border-left-color: var(--pst-color-background) !important;
+}
+
+/* Custom styles for Algolia DocSearch v3 items and modals */
+/* to make them congruous with PST on light and dark themes */
+
+@import url("https://cdn.jsdelivr.net/npm/@docsearch/css@3");
```
arjxn-py, line 144: Afaik all `@import` should be on top in a css, is it not giving any issue because it's not?

template/docs/_static/pybamm.css
```diff
@@ -0,0 +1,230 @@
...
+.sd-card {
+  background: #fff;
+  border-radius: 0;
+  padding: 30px 10px 20px 10px;
+  margin: 10px 0px;
+}
+
+.sd-card .sd-card-header {
+  text-align: center;
+}
```
arjxn-py, line 52:
```suggestion
```

template/docs/_static/pybamm.css
```diff
@@ -0,0 +1,230 @@
...
+
+.sd-card .sd-card-img-top {
+  height: 52px;
+  width: 52px;
+  margin-left: auto;
+  margin-right: auto;
+}
+
+.sd-card .sd-card-header {
+  border: none;
```
arjxn-py, line 66:
```suggestion
  text-align: center;
  border: none;
```

template/docs/conf.py.jinja
```diff
@@ -58,6 +58,13 @@ html_last_updated_fmt = "%Y-%m-%d"
 html_static_path = ["_static"]
 html_file_suffix = ".html"
 
+# For edit button
+html_context = {
+    "github_user": "{{org}}",
```
arjxn-py, line 63: org?

arjxn-py (changes requested):
Thanks @santacodes, some comments below

template/docs/_static/pybamm.css
santacodes: I just imported this CSS from pybamm just to fill in the `_static` directory discussed [here](https://github.com/pybamm-team/pybamm-cookiecutter/pull/39#discussion_r1704193154), which was failing the doctests. Without any content inside the directory, it wouldn't copy and fail the tests again so Ferran just added a quick fix by adding the CSS file so it would copy and not fail the doctests. The file is unnecessary and we could maybe replace it with something else to have something inside the `_static` directory.

arjxn-py: it's not on top, but never mind as this file have been copied from pybamm as it is.

arjxn-py (approved):
Looks good

santacodes:
@agriyakhetarpal do you have any ideas on what we could replace the `pybamm.css` file with in the `_static` folder? I think it would be good to merge after that. Refer [here](https://github.com/pybamm-team/pybamm-cookiecutter/pull/39#discussion_r1704193154).

.github/workflows/test_on_push.yml
```diff
@@ -134,3 +134,39 @@ jobs:
...
+        uses: yezz123/setup-uv@v4
+        with:
+          uv-venv: ".venv"
+
+      - name: Install nox
+        uses: wntrblm/nox@2024.04.15
+
+      - name: Install copier and jinja2_time and generate a template with default values
+        run: |
+          uv pip install copier jinja2_time
```
agriyakhetarpal, line 164:
```suggestion
          uv pip install copier jinja2-time
```

agriyakhetarpal (changes requested):
Thanks, @santacodes! I think it is better to remove the CSS file altogether and keep `html_css_files = ["custom.css"]` commented out in `conf.py` since the PyData Sphinx Theme already provides its own styles. The CSS file contains a lot of PyBaMM-specific configuration that users can copy manually if they wish to adjust and tinker with things. This is the only change I would request – the rest of the code looks good to me, and I'm glad to approve this PR after that.

santacodes:
> Thanks, @santacodes! I think it is better to remove the CSS file altogether and keep `html_css_files = ["custom.css"]` commented out in `conf.py` since the PyData Sphinx Theme already provides its own styles. The CSS file contains a lot of PyBaMM-specific configuration that users can copy manually if they wish to adjust and tinker with things. This is the only change I would request – the rest of the code looks good to me, and I'm glad to approve this PR after that.

The doctests would fail without something inside the `_static` folder, as we cannot copy an empty directory while generating a project with copier. Is there something else we could add in the `_static` folder just as a dummy to keep the tests running or is there a way to suppress the warning for that?

agriyakhetarpal:
Oh, I see. In that case, we can add an empty `.gitkeep` file inside the `_static` folder, which is likely what you need.

agriyakhetarpal (approved):
Thanks for the fixes, @santacodes!

PR #44, opened 2024-08-14
Added versioning support for git archives

### LINKED ISSUE #42: Add versioning support for git archives
Given that we are using `hatch-vcs` (in both the project and the template), the following two files should be added in to support versioning from VCS for a user installing the project through a git archive -

- `.git_archival.txt` ([example](https://github.com/scikit-hep/vector/blob/main/.git_archival.txt) (can be copied entirely))
- `.gitattributes` ([example](https://github.com/scikit-hep/vector/blob/main/.gitattributes) (can be copied entirely))

santacodes:
Fixes #42 

Added `.gitattributes.txt` and `git_archival.txt` to both project and template.


arjxn-py (approved):

template/{% if backend == 'hatch' and vcs == true %}.git_archival.txt{% endif %}.jinja
```diff
@@ -0,0 +1,4 @@
+node: $Format:%H$
```
Saransh-cpp, line 1: Should be present on `vcs == true` and any backend.

template/{% if backend == 'hatch' and vcs == true %}.gitattributes.txt{% endif %}.jinja
```diff
@@ -0,0 +1 @@
+.git_archival.txt export-subst
```
Saransh-cpp, line 1: Ditto

Saransh-cpp (changes requested):
Thanks, @santacodes! A couple of comments below -

Saransh-cpp (approved):
Thanks!

PR #48, opened 2024-08-19
Added release workflow

santacodes:
Added `release.yml` file referred from [PyPI OIDC docs](https://docs.pypi.org/trusted-publishers/).


agriyakhetarpal (changes requested):
This isn't correct – the PyPI docs are likely truncated and display only a truncated part of what is needed to be added. You'll need to build an sdist and a wheel in one job using `pypa/build` (I like https://github.com/hynek/build-and-inspect-python-package), upload their artifacts, download them in another job, and and trigger the publishing from that job.

Also, you might have done this already, but please ensure that the name of the workflow file matches that set in the PyPI settings for `pybamm-cookie` that you might have set up by now (if you have, please feel free to add us there!) and that the name of the environment you added matches the one in the `environment:` section here, i.e., `release`.

santacodes:
@agriyakhetarpal @Saransh-cpp sorry for the delay, could you verify if the release workflow is now valid? `pypa/gh-action-pypi-publish@release/v1` action by default looks for the `sdist` and `whl` from the `dist/` directory so a simple `python -m build` should do it I guess?

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+          uv-venv: ".venv"
+
+      - name: Install dependencies
+        run: uv pip install build hatch
+
+      - name: Build package
+        run: python -m build
+
+      - name: Publish package distributions to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
```
agriyakhetarpal, line 30:
```suggestion
      - name: Publish package distributions to PyPI
        uses: pypa/gh-action-pypi-publish@ec4db0b4ddc65acdf4bff5fa45ac92d78b56bdf0  # v1.9.0
```

agriyakhetarpal (changes requested):
No, this still isn't valid, but the structure is getting towards the right place!
- The idea is to build the sdist and wheel in an entirely separate job, upload them as artifacts (you're right – both can be kept in a `dist/` folder).
- We don't need to set up `uv`, GHA runners come with `pipx` by default, so you can just run `pipx run build` instead of `python -m build`.
- Then, in a separate Ubuntu job, download the artifacts, and grant just that job `id-token: write` permissions – the separation of the jobs is important for security reasons: please read the end of the [Trusted publishing](https://github.com/pypa/gh-action-pypi-publish/tree/release/v1/?tab=readme-ov-file#trusted-publishing) section of the documentation. 

Also, the latest release of `pypa/gh-action-pypi-publish` is v1.9.0, could you update that and set it to point to the hash of the v1.9.0 commit? I'll suggest a change below for it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
+jobs:
+  build-wheels:
```
agriyakhetarpal, line 2:
```suggestion
  build:
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+    name: upload pybamm-cookie wheel artifacts
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout pybamm-cookie
+      - uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.12
```
agriyakhetarpal, line 12:
```suggestion
```

We'll need to keep updating the Python version, so we might as well remove it and let it use the latest available one

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Build package
+        run: pipx run build
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v4.3.6
+        with:
+          name: pybamm_cookie_wheel
+          path: ./dist/*.whl
+          if-no-files-found: error
```
agriyakhetarpal, line 22:
```suggestion
          name: sdist_and_wheel
          path: ./dist/*
          if-no-files-found: error
```

`pipx run build` creates both the sdist and the wheel, so why not upload both?

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+
+  publish-pypi:
+    name: Upload package to PyPI
+    needs: [build-wheels]
```
agriyakhetarpal, line 26:
```suggestion
    needs: [build]
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: ubuntu-latest
+    environment: release
+    permissions:
+      id-token: write
+
+    steps:
+      - name: Download all artifacts
+        uses: actions/download-artifact@v4
+        with:
+          path: artifacts
```
agriyakhetarpal, line 36:
```suggestion
        with:
          path: artifacts
          merge-multiple: true
```

This change will cause the artifacts to get downloaded into the `artifacts/` folder, and not inside `artifacts/sdist_and_wheel/`, which should be better.

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
+jobs:
+  build-wheels:
+    name: upload pybamm-cookie wheel artifacts
```
agriyakhetarpal, line 3:
```suggestion
    name: Build and pybamm-cookie sdist and wheel
```

.github/workflows/release.yml
agriyakhetarpal, line 26:
```suggestion
    needs: [build]
```

santacodes: I guess this suggestion got duplicated

agriyakhetarpal: Oops, thanks!

agriyakhetarpal (approved):

agriyakhetarpal:
Thanks for the fixes, @santacodes! Let's create the release in the meeting in a short while from now.

PR #62, opened 2024-09-08
Fix coverage session inside generated project

### LINKED ISSUE #61: Coverage not working
I cloned the project recently and the coverage was not working as the action did not find any file to push. I found a way to fix it (probably not ideal, see https://github.com/mmsg-warwick/encapsulated-LTES/pull/4).

santacodes:
The coverage session was the same as the fix Ferran mentioned in this [commit](https://github.com/pybamm-team/pybamm-cookie/commit/5afcceeccd157aa12177a53b834748da7c39e446), reverting to this.
Fixes #61

### DIFF
--- a/template/{{ project_name }}/noxfile.py.jinja
+++ b/template/{{ project_name }}/noxfile.py.jinja
@@ -56,9 +56,10 @@ def run_user_tests(session):
 @nox.session(name="coverage")
 def run_coverage(session):
     """Run the coverage tests and generate an XML report."""
-    session.posargs.append("--cov=src/{{project_slug}}")
-    session.posargs.append("--cov-report=xml")
-    run_user_tests(session)
+    session.install("setuptools", silent=False)
+    session.install("coverage", silent=False)
+    session.install("-e", ".[dev]", silent=False)
+    session.run("pytest", "--cov=src/{{project_slug}}", "--cov-report=xml", "tests/")
 
 @nox.session(name="dev")
 def set_dev(session):


santacodes:
Also would this fix require a new release of version `v0.1.2`?

template/{{ project_name }}/noxfile.py.jinja
```diff
@@ -56,9 +56,10 @@ def run_user_tests(session):
 @nox.session(name="coverage")
 def run_coverage(session):
     """Run the coverage tests and generate an XML report."""
-    session.posargs.append("--cov=src/{{project_slug}}")
-    session.posargs.append("--cov-report=xml")
-    run_user_tests(session)
+    session.install("setuptools", silent=False)
```
agriyakhetarpal, line 59: Are we sure we need to install `setuptools` here? Does we receive failures without it?

agriyakhetarpal (approved):
Thanks, @santacodes! Yes, it would be nice to publish a v0.1.2 release.

template/{{ project_name }}/noxfile.py.jinja
santacodes: Oops I didn't notice this, we certainly don't need setuptools. I'll remove it.

santacodes: @agriyakhetarpal I think we do need setuptools, what if the generated project has a setuptools backend?

agriyakhetarpal: I see, we can keep it in that case :)

PR #119, opened 2025-05-24
Reduce generated project jobs to steps 

### LINKED ISSUE #117: Don't use too many CI jobs
I don't exactly recall why we chose to run this [big array of CI jobs](https://github.com/pybamm-team/pybamm-cookie/blob/2eead9bf7829a39ee465ebabab36b08c95f9fcc5/.github/workflows/test_on_push.yml#L111-L115) last year, but I think we could cut them down and still not increase CI time too significantly with our testing, and we could even reduce it for that matter. Right now, we have the overhead of our limited concurrency limit of 8 jobs (we're not on a paid plan from GitHub), the fact that the CI jobs check out the repository and the runners are set up and torn down, etc. 

If we were to remove the Python version, build backend, and vcs from matrix variables but rather move them into the job's steps, that would help. That is, we can keep reusing `actions/setup-python` in multiple steps within the same job, and we can move into different directories to run tests against setuptools/hatch + vcs/no-vcs within the same job. That would get us from 64 jobs to 4 jobs (one per platform), and as the rests barely take 15 seconds, an individual CI job would take 120 seconds, and that is surely fine.

santacodes:
The previous CI job used `64` jobs to test all the combinations of generated projects using `copier`; now the number of combinations has reduced to `16`. I think it would be better to keep python versions into their separate matrix as it would be much more cleaner and readable that way, personally I don't think it would be worth it to split them into their own steps with multiple python virtual environments, writing hacky scripts to generate and run tests on generated projects.

Fixes #117

### DIFF
--- a/.github/workflows/test_on_push.yml
+++ b/.github/workflows/test_on_push.yml
@@ -111,10 +111,8 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-13, macos-14, windows-latest]
         python-version: ["3.9", "3.10", "3.11", "3.12"]
-        backend: [hatch, setuptools]
-        vcs: [true, false]
 
-    name: Generated Project Tests (${{ matrix.os }} / Python ${{ matrix.python-version }} / Backend ${{ matrix.backend }} / VCS ${{ matrix.vcs }})
+    name: Generated Project Tests (${{ matrix.os }} / Python ${{ matrix.python-version }})
 
     steps:
       - name: Checkout pybamm-cookie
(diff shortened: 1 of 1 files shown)


agriyakhetarpal (commented):
Thanks, @santacodes! I see that there are two annotations for the CI run at https://github.com/pybamm-team/pybamm-cookie/actions/runs/15229557825?pr=119, could you also fix them? Perhaps it's just an extraneous input?

Also, I think we can reduce this even further. On the same link, you can see that we have 16 jobs (and therefore 16 runners) for the `template_test` matrix, and then we have tests for all of them in 16 jobs in `generated_project_tests`. Your idea of keeping Python versions in separate jobs is alright – I agree there. However, running the generated project tests as a step within `template_test` should reduce 16 jobs from the current 36.

PR #128, opened 2025-07-22
Tests the new entry points API in pybamm

medha-14:


medha-14:
Could iI get some help with the `test_parameter_sets_entry_point_load ` failure, not sure why the dicts aren’t matching anymore. Only difference I can see is "citations" vs "notcite", am i missing something? What could be causing this?

santacodes:
> Hmm, the only source of truth for `notcite` across both the PyBaMM and pybamm-cookie repositories is here:
> 
> https://github.com/pybamm-team/pybamm-cookie/blob/f4f2defd34d505efe3ddf9db128e6f8712202b74/template/%7B%7B%20project_name%20%7D%7D/src/%7B%7Bproject_slug%7D%7D/parameters/input/Chen2020.py#L370
> 
> Perhaps this can help? I don't recall the context in which this was added, as it has been a while. Perhaps @santacodes does 🤔

I don't exactly recall this @agriyakhetarpal, as they worked out of the box when I copied and pasted these parameter values here to test the entry points.

PR #171, opened 2026-03-26
[CI] Deprecate python3.9 and add 3.13 and 3.14 to test matrix

santacodes:
- As python 3.9 is reaching EOL, have deprecated it and added 3.13 and 3.14 to the test matrix. 
- Have made 3.13 the standard version for testing docs.

### DIFF
--- a/.github/workflows/test_on_push.yml
+++ b/.github/workflows/test_on_push.yml
@@ -36,7 +36,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-14, macos-15-intel, windows-latest]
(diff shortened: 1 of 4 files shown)


Saransh-cpp (changes requested):
Thanks, @santacodes! I think we should also update the template CI files + `pyproject.toml` for both the project and the template.

santacodes:
> Thanks, @santacodes! I think we should also update the template CI files + `pyproject.toml` for both the project and the template.

@Saransh-cpp have addressed them in the latest commit, perhaps we could also make a release after this?

Saransh-cpp (approved):
Thanks, @santacodes! Yes, a release (`v0.2.0`) sounds nice. Feel free to make one!

